PR #6, opened 2024-11-07
feat / add date formatting

vladjerca:
## API Changes:

- Extend the extendedQuerySchema to include new values and make it optional.
- Update episodeSchema and showSchema to include optional fields for new episode types and images.
- Update imports and response schemas to use the new showListResponseSchema.

## Client Changes:

- Enhance the Episode component to handle new episode types and display tags for premieres and finales.
- Add internationalization support for new episode types. 
- Update the upcoming episodes request to include new episode types and extended query parameters. 
- Integrated the EpisodeIntlProvider and updated the episode component usage.

## Misc Changes:

- Add translations for new episode types and associated episode tags.
- Add date-fns as a new dependency.
- Update the font family.
- Add locale mappings for date formatting.
- Removed the demo test file projects/client/src/demo.spec.ts.


projects/client/src/lib/requests/calendars/upcomingEpisodes.ts
```diff
@@ -43,19 +49,22 @@ export function upcomingEpisodes({
 
       return body
         .map((item) => {
-          const posterCandidate = item.episode.images.screenshot.at(0) ??
-            item.show.images.fanart.at(0);
+          const posterCandidate = item.episode.images!.screenshot.at(0) ??
```
seferturan, line 52: In the future we should add a placeholder or something for when there are no images.

vladjerca: @seferturan We're prepared:

```ts
export function prependHttps(
  url: string | undefined,
  placeholder?: string,
): string {
  if (!url) {
    return placeholder ?? '';
  }

  return url.startsWith('http') ? url : `https://${url}`;
}
```

PR #96, opened 2024-12-04
feat: add first version of the footer

seferturan:
## 🎶 Notes 🎶

- Adds the first version of the footer; it is still missing the links to pages (that we don't even have yet), and some social links.
- When the area becomes too small, the size is reduced and most elements are hidden.
- Also tested it in Firefox & Safari

## 👀 Examples 👀
Screenshot 2024-12-04 at 16 40 15

Screenshot 2024-12-04 at 16 40 53

Screenshot 2024-12-04 at 16 46 15

### DIFF
--- a/projects/client/src/lib/sections/navbar/Navbar.svelte
+++ b/projects/client/src/lib/sections/navbar/Navbar.svelte
@@ -39,57 +39,62 @@
   let inputElement: HTMLInputElement;
 </script>
 
-<nav class="trakt-navbar" class:trakt-navbar-scroll={isScrolled}>
-  <div class="trakt-logo">
-    <Link color="inherit" href="/">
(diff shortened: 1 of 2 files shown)


projects/client/src/lib/sections/footer/Footer.svelte
```diff
@@ -0,0 +1,23 @@
...
+
+<style>
+  .trakt-footer {
+    height: var(--ni-300);
+    margin-top: var(--ni-120);
+    padding-left: var(--layout-distance-side);
+    padding-right: var(--layout-distance-side);
+
+    /* TODO: remove when RenderFor re-render is fixed */
+    @media (max-width: 768px) {
```
vladjerca, line 19: Not sure I get this, but alright

PR #112, opened 2024-12-09
feat: add landing section

seferturan:
## 🎶 Notes 🎶
- Added `@sveltejs/enhanced-img` as a dependency to auto-optimize images.
- Extracted the movie and show summary mappers.
- Added API support for getting the trending items.
- Removed `font-variant-numeric` since it caused rendering issues for certain numbers.
- Added the first version of the landing section. Layout is not 100% final yet.
- The resize issue caused by `RenderFor` will be fixed with this PR: https://github.com/trakt/trakt-lite/pull/110

## 🔮 Follow ups 🔮
- The third show cover will be replaced with a random episode from one of the trending shows.
  - The layout will be updated when adding this.
- The `TrendingMedia` component footer is empty until we have a CTA to place there.
- The `Phones` component will be replaced when we have the new assets.

## 👀 Example 👀
Screenshot 2024-12-09 at 20 33 31

### DIFF
--- a/projects/client/src/lib/requests/models/MediaSummary.ts
+++ b/projects/client/src/lib/requests/models/MediaSummary.ts
@@ -1,0 +1,18 @@
+import type { Genre } from '$lib/api.ts';
+
+export type MediaSummary = {
+  id: number;
+  slug: string;
+  runtime: number;
+  title: string;
+  tagline: string;
+  poster: {
+    url: string;
+  };
+  cover: {
+    url: string;
+  };
+  genres: Genre[];
+  overview: string;
+  trailer?: string;
+};
--- a/projects/client/src/lib/requests/models/MovieSummary.ts
+++ b/projects/client/src/lib/requests/models/MovieSummary.ts
@@ -1,18 +1,3 @@
-import type { Genre } from '$lib/api.ts';
+import type { MediaSummary } from './MediaSummary.ts';
 
-export type MovieSummary = {
-  id: number;
-  slug: string;
-  runtime: number;
-  title: string;
-  tagline: string;
-  poster: {
-    url: string;
-  };
-  cover: {
-    url: string;
-  };
-  genres: Genre[];
-  overview: string;
-  trailer?: string;
-};
+export type MovieSummary = MediaSummary;
--- a/projects/client/src/lib/requests/models/ShowSummary.ts
+++ b/projects/client/src/lib/requests/models/ShowSummary.ts
@@ -1,18 +1,3 @@
-import type { Genre } from '$lib/api.ts';
+import type { MediaSummary } from './MediaSummary.ts';
 
-export type ShowSummary = {
-  id: number;
-  slug: string;
-  runtime: number;
-  title: string;
-  tagline: string;
-  poster: {
-    url: string;
-  };
-  cover: {
-    url: string;
-  };
-  genres: Genre[];
-  overview: string;
-  trailer?: string;
-};
+export type ShowSummary = MediaSummary;


projects/client/src/lib/sections/landing/components/Phones.svelte
```diff
@@ -0,0 +1,37 @@
...
+    src={androidScreenshot}
+    class="trakt-background-phone"
+    alt="Android screenshot"
+  />
+  <enhanced:img
+    src={iosScreenshot}
+    class="trakt-foreground-phone"
+    alt="iOS screenshot"
+  />
+</div>
```
vladjerca, line 21: I think we need to configure compression. Images are very blurry (and some artifacting is visible). We need to find a good sweet spot between quality and size.

image

projects/client/src/lib/requests/_internal/mapMovieResponseToMovieSummary.ts
```diff
@@ -0,0 +1,54 @@
...
+  poster: {
+    url: string;
+  };
+  cover: {
+    url: string;
+  };
+  genres: Genre[];
+  overview: string;
+  trailer?: string;
+};
```
vladjerca: Maybe we should have a public type called `MovieSummary`? Other mappers can simply return this common object rather than renaming them.

projects/client/src/lib/requests/queries/movies/movieSummaryQuery.ts
```diff
@@ -1,59 +1,13 @@
...
-    url: string;
-  };
-  cover: {
-    url: string;
-  };
-  genres: Genre[];
-  overview: string;
-  trailer?: string;
-};
+export type MovieSummary = InternalMovieSummary;
```
vladjerca: Then this can just be the public `MovieSummary`

projects/client/src/lib/requests/queries/movies/movieTrendingQuery.ts
```diff
@@ -0,0 +1,48 @@
+import type { MovieTrendingResponse } from '$lib/api.ts';
+import { api, type ApiParams } from '../../_internal/api.ts';
+import {
+  type InternalMovieSummary,
+  mapMovieResponseToMovieSummary,
+} from '../../_internal/mapMovieResponseToMovieSummary.ts';
+
+export type TrendingMovie = InternalMovieSummary & {
+  watchers: number;
+};
```
vladjerca: And this can be `TrendingMovie = MovieSummary & ...`

projects/client/src/lib/requests/queries/shows/_internal/mapShowResponseToShowSummary.ts
```diff
@@ -0,0 +1,49 @@
...
+  poster: {
+    url: string;
+  };
+  cover: {
+    url: string;
+  };
+  genres: Genre[];
+  overview: string;
+  trailer?: string;
+};
```
vladjerca: Same as with `Movie` I believe we already have an example of this with the `MediaSummary` this might even be a duplicate of that actually.

projects/client/src/lib/sections/landing/StepsInt.ts
```diff
@@ -0,0 +1,10 @@
+export enum LandingStepType {
+  Track = 'track',
+  Discover = 'discover',
+  Share = 'share',
+}
+
+export type StepsIntl = {
+  label: (step: LandingStepType) => string;
+  description: (step: LandingStepType) => string;
+};
```
vladjerca: Filename is wrong: `StepsInt`

projects/client/src/lib/sections/landing/useTrendingItems.ts
```diff
@@ -0,0 +1,60 @@
...
+  if (!items) {
+    return [];
+  }
+
+  const randomSort = () => Math.random() - 0.5;
+
+  return items
+    .toSorted(randomSort)
+    .slice(0, count ?? items.length);
+}
```
vladjerca: Please extract utility functions:

- `shuffle(items: T[])`
- `random(min, max)`

projects/client/src/lib/sections/landing/useTrendingItems.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  const trendingMovies = createQuery(
+    movieTrendingQuery(),
+  );
+
+  return {
+    shows: derived(trendingShows, ($shows) => {
+      return $shows.isFetching
+        ? []
+        : getRandomItems($shows.data, RANDOM_SHOW_COUNT);
```
vladjerca: `shuffle(data).slice(0, RANDOM_SHOW_COUNT)`

projects/client/src/lib/sections/landing/useTrendingItems.ts
```diff
@@ -0,0 +1,60 @@
...
+    }),
+    //TODO replace with episode
+    show: derived(trendingShows, ($shows) => {
+      return $shows.isFetching ? undefined : getRandomItem($shows.data);
```
vladjerca: `shuffle(data).at(0)`

projects/client/src/lib/sections/landing/useTrendingItems.ts
```diff
@@ -0,0 +1,60 @@
...
+    }),
+    movie: derived(trendingMovies, ($movies) => {
+      return $movies.isFetching ? undefined : getRandomItem($movies.data);
```
vladjerca: shuffle(data).at(0)

projects/client/vite.config.ts
```diff
@@ -55,6 +56,7 @@ export default defineConfig(({ mode }) => ({
     }),
     denoSveltekitExit(),
     svelteTesting(),
+    enhancedImages(),
```
vladjerca, line 65: Configure for better compression? Or is it the source image itself that's weird? Maybe we could ask @anodpixels for a better export for source?

projects/client/src/lib/sections/landing/components/TrendingItems.svelte
```diff
@@ -0,0 +1,58 @@
...
+        }
+      }
+    }
+
+    .trakt-landing-phones {
+      z-index: 1;
+
+      @container (max-width: 768px) {
+        display: none;
+      }
```
vladjerca: I don't think this is a wise use-case. @container queries are useful for layout out stuff but combining with display creates weird situations:

`display: none` changes the size which will basically guarantee that the content will never re-appear unless hidden. This would probably work in scenarios where the parent has a with and the container is `100%` relative to the parent. But other than that this will end up with weird scenarios:

image

image

projects/client/src/lib/sections/landing/components/Phones.svelte
seferturan: I added the new screenshots for this. Note: these are still not the final screenshots, will revisit when also adding the phones.

projects/client/src/lib/requests/_internal/mapMovieResponseToMovieSummary.ts
seferturan: Done 👌

projects/client/src/lib/requests/queries/shows/_internal/mapShowResponseToShowSummary.ts
seferturan: Done 👌

projects/client/src/lib/sections/landing/StepsInt.ts
seferturan: I renamed the file.

projects/client/src/lib/sections/landing/useTrendingItems.ts
seferturan: Done 👌

projects/client/vite.config.ts
seferturan: Default config seems to be fine, the source images were crap.

projects/client/src/lib/sections/landing/components/TrendingItems.svelte
seferturan: "Fixed". Only enabled it for desktops until we have a responsive design.

projects/client/src/lib/sections/landing/useTrendingItems.ts
```diff
@@ -0,0 +1,40 @@
...
+      return $shows.isFetching || !$shows.data
+        ? undefined
+        : shuffle($shows.data).at(0);
+    }),
+    movie: derived(trendingMovies, ($movies) => {
+      return $movies.isFetching || !$movies.data
+        ? undefined
+        : shuffle($movies.data).at(0);
+    }),
+  };
```
vladjerca, line 39: Refactor in follow-up please `data ?? []`

PR #225, opened 2025-01-05
feat(pwa): blend in with top bar

vladjerca:
@seferturan please look at this as is

BEFORE:

AFTER:


vladjerca:
> Looks and works great on Android :D I'd say let's do it.

Mark it as ready for review.

PR #236, opened 2025-01-08
feat: take media item statuses into count

seferturan:
## 🎶 Notes 🎶
- Fixes the issue reported here: https://github.com/trakt/trakt-lite/issues/223
- The coming soon section now takes statuses into count. Movies that are canceled will no longer show up.
  - Only movies with a status that would indicate it is actually coming is taken into count:
    - i.e.: `planned`, `post production`, `in production` & `upcoming`
  - This also filters out movies that were released, but have no known release date.
- This also fixed a related issue in the media details. When there is no usable premiere date, the status will be shown.
- Also updates the label of the premiere date if it's in the future.

## 👀 Examples 👀
**Show only relevant upcoming movies**
Before:
Screenshot 2025-01-08 at 13 58 55

After:
Screenshot 2025-01-08 at 13 59 06

**Showing status instead of date**
Before:
Screenshot 2025-01-08 at 14 00 10

After:
Screenshot 2025-01-07 at 20 54 15

**Premiere date example**
Screenshot 2025-01-07 at 21 26 20

### DIFF
--- a/projects/client/src/lib/sections/lists/stores/useComingSoon.ts
+++ b/projects/client/src/lib/sections/lists/stores/useComingSoon.ts
@@ -1,8 +1,16 @@
 import type { MediaType } from '$lib/models/MediaType.ts';
+import type { MediaStatus } from '$lib/requests/models/MediaStatus';
 import {
   useWatchlistList,
 } from '$lib/sections/lists/stores/useWatchlistList.ts';
 import { derived } from 'svelte/store';
+
+const IN_PROGRESS_STATUSES: MediaStatus[] = [
+  'planned',
+  'post production',
+  'in production',
+  'upcoming',
+] as const;
 
 export function useComingSoon(type: MediaType) {
   const { list: watchlist, isLoading } = useWatchlistList({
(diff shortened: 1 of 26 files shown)


projects/client/src/lib/sections/lists/stores/useComingSoon.ts
```diff
@@ -4,6 +4,13 @@ import {
 } from '$lib/sections/lists/stores/useWatchlistList.ts';
 import { derived } from 'svelte/store';
 
+const IN_PROGRESS_STATUSES = [
```
vladjerca: Can we type this to make sure we don't use invalid string types?

projects/client/src/lib/sections/summary/components/MediaDetails.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+import { MovieHereticStudiosMappedMock } from '$mocks/data/summary/movies/heretic/mapped/MovieHereticStudiosMappedMock';
+import MediaDetails, { type MediaDetailsProps } from './MediaDetails.svelte';
+
+import { render, screen } from '@testing-library/svelte';
+import { describe, expect, it } from 'vitest';
+import type { MediaSummary } from './MediaSummary';
+
+describe('MediaDetails', () => {
+  const defaultProps: MediaDetailsProps = {
+    //See TODO in MovieHereticMappedMock.ts for why this is cast
```
vladjerca, line 13: I might have an idea here, let's talk tomorrow about this

vladjerca (changes requested):

projects/client/src/lib/sections/lists/stores/useComingSoon.ts
seferturan: Done 👌

PR #240, opened 2025-01-09
feat: use gateway for requests

seferturan:
## 🎶 Notes 🎶

- Cookie content is now encrypted.
- Auth headers are removed; it is now handled via a gateway.
- Also fixes a small issue where logging out would still refetch active querires.


seferturan:
Drafted; gonna add a bit to migrate existing sessions so people won't be signed out once it's deployed.

seferturan:
Added a step to migrated the serialized cookies:

https://github.com/user-attachments/assets/3536f4bb-d58d-41b6-990a-7d627fa45e4d

vladjerca (approved):
LGTM: https://youtu.be/q6GwSzGQsHk?feature=shared&t=24

PR #287, opened 2025-01-14
chore: use perfective aspect for 'Recently Watched' messages in Polish

Jackenmen:
Small, perhaps pedantic, thing but the inconsistent verb aspect can be seen in different messages on a single (profile) page which makes this a bit worse imo.


seferturan:
@Jackenmen, can you please squash them into a single commit, or rename the `one more` one? 🙏

PR #292, opened 2025-01-14
feat: add summary page for people

seferturan:
## 🎶 Notes 🎶

- Adds summary page for people.
- You can click on an actor on a show/movie summary page to get there.
- Will get follow-ups:
  - Default to a placeholder background image when there is none.
    - The background image of the item you came from is retained; e.g. if you click on an actor from the `Heretic` page, it will keep that background image.
  - Add a list with `known for`/`most popular` shows/movies that actor has been in.

## 👀 Example 👀

https://github.com/user-attachments/assets/083d0fd9-d5f3-4d35-88f2-a8f337df5a90

### DIFF
--- a/projects/client/src/lib/sections/summary/PeopleSummary.svelte
+++ b/projects/client/src/lib/sections/summary/PeopleSummary.svelte
@@ -1,0 +1,12 @@
+<script lang="ts">
+  import type { PersonSummary } from "$lib/requests/models/PersonSummary";
+  import ActorSummary from "./components/ActorSummary.svelte";
+
+  const {
+    person,
+  }: {
+    person: PersonSummary;
+  } = $props();
+</script>
+
+<ActorSummary {person} />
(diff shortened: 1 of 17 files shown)


projects/client/src/lib/sections/summary/PeopleSummary.svelte
```diff
@@ -0,0 +1,12 @@
...
+  import ActorSummary from "./components/ActorSummary.svelte";
+
+  const {
+    person,
+  }: {
+    person: PersonSummary;
+  } = $props();
+</script>
+
+<ActorSummary {person} />
```
vladjerca, line 12: This feels like a needless component 🤔 

_(unless you have a super secret future plan for it)_

seferturan: I do indeed:
image

vladjerca (approved):

PR #296, opened 2025-01-14
fix: remove non-scopeable selectors

seferturan:
## 🎶 Notes 🎶

- I'll need to look into this a bit more as a follow up. We have more cases where we use global in the root of style tags. But not all of them have this issue 🤔

## 👀 Example 👀
Fixes these:
Screenshot 2025-01-14 at 22 34 38

### DIFF
--- a/projects/client/src/lib/sections/lists/SeasonList.svelte
+++ b/projects/client/src/lib/sections/lists/SeasonList.svelte
@@ -92,11 +92,3 @@
     {/if}
   {/snippet}
 </ShadowList>
-
-<style>
-  :global(.shadow-list-container) {
-    :global(.trakt-dropdown-list-container[data-size="small"]) {
-      margin-right: var(--ni-neg-36);
-    }
-  }
-</style>
--- a/projects/client/src/lib/sections/summary/components/MediaOverview.svelte
+++ b/projects/client/src/lib/sections/summary/components/MediaOverview.svelte
@@ -11,15 +11,6 @@
   const { title, overview }: MediaOverviewProps = $props();
 </script>
 
-<ClampedText
-  classList="trakt-media-overview secondary"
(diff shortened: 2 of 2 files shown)


projects/client/src/lib/sections/lists/SeasonList.svelte
```diff
@@ -92,11 +92,3 @@
...
   {/snippet}
 </ShadowList>
-
-<style>
-  :global(.shadow-list-container) {
-    :global(.trakt-dropdown-list-container[data-size="small"]) {
-      margin-right: var(--ni-neg-36);
-    }
-  }
-</style>
```
vladjerca, line 102: @seferturan this changes spacing and UX, class should be here.

That warning is added by the sveltekit compiler since the last upgrade saying we should move these global overrides to `.css` files and do `import './style.css`.

vladjerca (changes requested):

projects/client/src/lib/sections/lists/SeasonList.svelte
vladjerca: To clarify:

```svelte
<style>
    :global(.selector) {
       /* This results in warning. */
    }
</style>

<style>
    .parent {
        :global(.selector) {
           /* This is considered okay since it's namespaced. */
        }
    }
</style>
```

To be honest if you ask me this is technically an "optimization" for the dev-server to be "faster". I would see if we could silence this warning TBH, I find it easier to follow the styles alongside the component (even if applied globally).

What are your thoughts @seferturan?

seferturan:
Drafting it until we have a better understanding.
Current info:

- These styles are not applied at all.
- The warning was only for those 2 components; however, all other global selectors in the style root of other components are also not applied. But those don't seem to give a warning...

PR #303, opened 2025-01-16
feat / spoiler free prototype

vladjerca:
## Spoiler-Free Viewing: Protecting the Sanctity of Unspoiled Media

This pull request introduces a new feature that allows users to enjoy their media experience without fear of spoilers.

### User Settings Update

- Added an `isSpoilerFree` setting to the `UserSettings` type, empowering users to control their exposure to spoilers.

### Component Updates

- The `EpisodeItem.svelte`, `NextEpisodeItem.svelte`, and `UpcomingEpisodeItem.svelte` components now respect the `isSpoilerFree` setting, conditionally blurring episode covers and titles to prevent spoilers.
- The `MediaOverview.svelte` component also supports the spoiler-free setting, ensuring that sensitive plot details remain hidden from unsuspecting eyes.

### New Utility

- Created a `useSpoiler` store to provide a centralized and reactive way to access the user's spoiler-free preference.

### CSS Changes

- Added a `.spoiler` class to apply a blur filter to images and text, providing a visual cue for spoiler-free content.

(This new feature, like a vigilant censor guarding the gates of revelation, protects the sanctity of unspoiled media, allowing users to explore the world of movies and TV shows without fear of encountering unwanted plot twists or character deaths. The spoiler-free setting, a sanctuary for those who cherish the thrill of discovery, ensures that every reveal, every twist, and every dramatic climax is experienced as intended, unsullied by the taint of prior knowledge.)

image
image


projects/client/src/lib/sections/lists/components/EpisodeItem.svelte
```diff
@@ -34,6 +35,12 @@
     );
 
   const isFuture = $derived(episode.airDate > new Date());
+  const { isSpoilerFree } = useSpoiler();
+
+  const src = $derived(
+    ($isSpoilerFree ? show.cover.url.thumb : episode.cover.url) ??
+      show.cover.url.thumb,
+  );
```
tysonkerridge: This might be a nitpick but would it make more sense for the logic to be like below for the case where the show image is null and no need to check it again:
```svelte
const src = $derived(
  $isSpoilerFree
    ? show.cover.url.thumb
    : (episode.cover.url ?? show.cover.url.thumb),
);
  ```
  
And I'm unfamiliar with svelte but would it make sense to create a helper function to use for the various spots. This also means if any logic is extended (such as revealing if it's been watched or the user manually reveals it by clicking, etc, it can be added here too), eg. 
```svelte
// optional syntax might be wrong here
export function getImageUrl(
  isSpoilerFree: boolean,
  showImage: string,
  episodeImage?: string,
): string {
  return isSpoilerFree ? showImage : (episodeImage ?? showImage);
}
```

projects/client/src/lib/sections/summary/components/MediaOverview.svelte
```diff
@@ -2,17 +2,19 @@
...
     overview: string;
   };
 
   const { title, overview }: MediaOverviewProps = $props();
+  const { isSpoilerFree } = useSpoiler();
 </script>
 
 <ClampedText
-  classList="trakt-media-overview secondary"
+  classList="trakt-media-overview secondary {$isSpoilerFree ? 'spoiler' : ''}"
```
tysonkerridge: This might need to be applied to `MediaTitle.svelte` but only for episode titles, not Movie/Show/Season I guess?

projects/client/src/lib/features/auth/queries/currentUserSettingsQuery.ts
```diff
@@ -83,6 +84,7 @@ function mapUserSettingsResponse(response: SettingsResponse): UserSettings {
           direction: browsing?.progress.on_deck.sort_how,
         },
       },
+      isSpoilerFree: Object.values(browsing?.spoilers ?? {}).some(Boolean),
```
tysonkerridge: I can't tell how this is actually getting the value but cool 👀

tysonkerridge (commented):
🙌 

Obviously it adds a bit of complexity but an extension to this is that once the episode is watched it reveals spoilers such as the title at least on the main page (though imo the up next show image being prioritised in general is nice to keep as mentioned previously for visual familiarity when viewing the sections rather than needing to read the titles). Having it always considered hidden/blurred on the Up next though is nice for when rewatching. 🙈

vladjerca:
> 🙌
> 
> Obviously it adds a bit of complexity but an extension to this is that once the episode is watched it reveals spoilers such as the title at least on the main page (though imo the up next show image being prioritised in general is nice to keep as mentioned previously for visual familiarity when viewing the sections rather than needing to read the titles). Having it always considered hidden/blurred on the Up next though is nice for when rewatching. 🙈

Don't make me close the PR 😛

projects/client/src/lib/sections/lists/components/EpisodeItem.svelte
vladjerca: The PR is Draft, I added reviewers for awareness... not for review

tysonkerridge: Tag me instead?

tysonkerridge:
> Don't make me close the PR 😛

If anything it was for awareness of additional functionality to consider, hence extension 🤷

PR #316, opened 2025-01-20
feat(landing): use assets with phone backgrounds

seferturan:
## 🎶 Notes 🎶

- Replace landing assets

## 👀 Example 👀
Screenshot 2025-01-20 at 13 10 05


vladjerca (changes requested):
These device outlines look horrible, we need better quality renders 🥲 

cc: @anodpixels @seferturan

seferturan:
With different phone assets:
Screenshot 2025-01-20 at 14 40 51

We can fine tune/change them later. For now, we can start with these so we can at least have some phones :P

vladjerca (approved):

PR #348, opened 2025-01-23
feat(watchnow): when no favorite service use first streaming service

seferturan:
## 🎶 Notes 🎶

- When there is no favorite streaming service found, we'll use the first available streaming service as the direct link service. These are sorted on prio, so for now it's good enough. We can re-visit, and see if we want a more complicated heuristic (or include on demand services also) later on.


projects/client/src/lib/components/buttons/watch-now/WatchNowButton.svelte
```diff
@@ -32,8 +32,13 @@
    */
 </script>
 
-{#if favoriteService}
-  <WatchNowDirectLinkButton {mediaTitle} {favoriteService} {i18n} {isLoading} />
+{#if directLinkService}
+  <WatchNowDirectLinkButton
```
vladjerca: `WatchNowLink`

projects/client/src/lib/components/buttons/watch-now/WatchNowButton.svelte
```diff
@@ -32,8 +32,13 @@
...
+    {mediaTitle}
+    {directLinkService}
+    {i18n}
+    {isLoading}
+  />
 {:else}
   <WatchNowDropdownButton {mediaTitle} {services} {i18n} {isLoading} />
```
vladjerca: `WatchNowDropdown`

projects/client/src/lib/components/buttons/watch-now/WatchNowButton.svelte
```diff
@@ -11,7 +11,7 @@
   type WatchNowButtonProps = {
     isLoading: boolean;
     mediaTitle: string;
-    favoriteService?: WatchNowStreaming;
+    directLinkService?: WatchNowStreaming;
```
vladjerca: `preferred`

vladjerca (approved):

PR #374, opened 2025-01-27
feat: rating actions

seferturan:
## 🎶 Notes 🎶

- Will need a responsiveness follow-up when the watch now dropdown is removed.

## 👀 Example 👀
Screenshot 2025-01-27 at 14 17 22


projects/client/src/lib/sections/summary/components/rating/_internal/RateActionButton.svelte
```diff
@@ -0,0 +1,41 @@
...
+
+  const style = $derived(isCurrentRating ? "flat" : "ghost");
+  const variant = $derived(isCurrentRating ? "primary" : "secondary");
+
+  const backgroundColor = $derived(
+    rating === SimpleRating.Great ? "var(--red-900)" : "var(--shade-10)",
+  );
+  const foregroundColor = $derived(
+    rating === SimpleRating.Great ? "var(--red-500)" : "var(--shade-500)",
+  );
```
vladjerca: @seferturan can we please replace this with a color lookup?

```ts
const RatingBackground = {
 [SimpleRating.Great]: 'var(--red-900)',
 ...
}

const RatingForeground = {
 [SimpleRating.Great]: 'var(--red-500)',
 ...
}
```

Better than ternaries and if's 🙏🏼

vladjerca (changes requested):

vladjerca (approved):

PR #399, opened 2025-02-01
feat(i18n): add Italian translations

iDavide94:
## Add Italian translations
This pull request is for the purpose of adding the Italian language to the app.

## Note on translation
The more correct translation of ***"Watchlist"*** is ***"Watchlist"***.

The [literal translation](https://www.wordreference.com/enit/watchlist) would be ***"lista di osservazione"*** or ***"lista di controllo"***, but not in the case of Trakt, because in Italian it does not reflect the actual use of Trakt's watchlist. A more accurate translation would be ***"lista delle cose che vorrei guardare"***, but more of a description than a translation. Therefore, I think it is more correct to keep the original term, as it is part of Trakt-specific terminology, and similar to other English terms in the Italian dictionary (e.g. ["blacklist"](https://www.treccani.it/vocabolario/black-list_(Neologismi)/)).


seferturan:
Thanks for the contribution :D
I applied a small fixup to fix the comma in the settings json.

PR #410, opened 2025-02-06
feat: popup menu

seferturan:
## 🎶 Notes 🎶

- Adds a popup menu
- Watch list and mark as watched actions now have a dropdown variant
- Media and summary cards use the popup menu

## 👀 Examples 👀

https://github.com/user-attachments/assets/fa15f8b1-d0a2-42ca-82ac-3a0396b93848

https://github.com/user-attachments/assets/51e28b4a-e7c3-4754-b32c-e9db47a1cc0c

### DIFF
--- a/projects/client/src/lib/components/buttons/popup/PopupMenu.svelte
+++ b/projects/client/src/lib/components/buttons/popup/PopupMenu.svelte
@@ -1,0 +1,147 @@
+<script lang="ts">
+  import MoreIcon from "$lib/components/icons/MoreIcon.svelte";
+  import { disableTransitionOn } from "$lib/utils/actions/disableTransitionOn";
+  import { slide } from "svelte/transition";
+  import type { PopupMenuProps } from "./PopupMenuProps";
+  import { usePortal } from "./_internal/usePortal";
+
+  const { items, ...props }: PopupMenuProps = $props();
+
+  const { portalTrigger, portal, isOpened } = usePortal();
+</script>
+
+<button
+  use:disableTransitionOn={"touch"}
+  use:portalTrigger
+  aria-haspopup="true"
+  class="trakt-popup-menu-button"
+  {...props}
+>
+  <MoreIcon />
+</button>
+
+{#if $isOpened}
+  <div
+    use:portal
+    class="trakt-popup-menu-container"
+    transition:slide={{ duration: 150 }}
+    onoutrostart={(e) => e.currentTarget.classList.add("removing")}
+  >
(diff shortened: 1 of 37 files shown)


projects/client/src/lib/components/buttons/popup/PopupMenu.svelte
```diff
@@ -0,0 +1,168 @@
...
+  }}
+>
+  <button
+    use:disableTransitionOn={"touch"}
+    use:portalHandler
+    aria-haspopup="true"
+    class="trakt-popup-menu-button"
+    onclick={(ev) => {
+      isPopupOpen.update((state) => !state);
+      onclick?.(ev);
```
vladjerca: Let's move this inside the `handler` and rename it ti `portalTrigger`

projects/client/src/lib/components/buttons/popup/PopupMenu.svelte
```diff
@@ -0,0 +1,168 @@
...
+  import MoreIcon from "$lib/components/icons/MoreIcon.svelte";
+  import { clickOutside } from "$lib/utils/actions/clickOutside";
+  import { disableTransitionOn } from "$lib/utils/actions/disableTransitionOn";
+  import { slide } from "svelte/transition";
+  import type { PopupMenuProps } from "./PopupMenuProps";
+  import { usePortalActions } from "./_internal/usePortalActions";
+
+  const { items, onclick, ...props }: PopupMenuProps = $props();
+
+  const { portalHandler, portal, isPopupOpen } = usePortalActions();
```
vladjerca: `usePortal` instead of `usePortalActions`

projects/client/src/lib/components/buttons/popup/PopupMenu.svelte
```diff
@@ -0,0 +1,168 @@
...
+{#if $isPopupOpen}
+  <div
+    use:portal
+    class="trakt-popup-menu-container"
+    transition:slide={{ duration: 150 }}
+    onoutrostart={(e) => e.currentTarget.classList.add("removing")}
+  >
+    <div class="trakt-popup-menu-button-placeholder"><MoreIcon /></div>
+    <div class="spacer"></div>
+    <ul onclickcapture={() => isPopupOpen.set(false)}>
```
vladjerca: this can also be set within the `portal` action so the `isPopupOpen` does not need to be exposed at all.

seferturan: Partially done, will follow up later. For now, it's returning a read-only `isOpen`.

PR #417, opened 2025-02-09
Fix/portal underlay

seferturan:
## 🎶 Notes 🎶

- Clickoutside action now also considers window scroll and resize as an outside click.
- Adds an underlay to the popup menu:
  - Prevents scrolling in lists when the menu is open.
  - Prevents actions when clicking outside to close it (see example).

## 👀 Examples 👀
Before:
https://github.com/user-attachments/assets/77500231-14b7-4e30-97cf-188dbf5b19f8

After:
https://github.com/user-attachments/assets/0b36e4df-d54d-4b0f-8859-bb3925252dca

### DIFF
--- a/projects/client/src/lib/components/buttons/popup/_internal/constants.ts
+++ b/projects/client/src/lib/components/buttons/popup/_internal/constants.ts
@@ -1,0 +1,1 @@
+export const PORTAL_UNDERLAY_ID = 'trakt-portal-underlay';
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -58,6 +58,7 @@ module.exports = {
         'profile',
         'pr_gate',
         'popular',
+        'popup',
         'pwa',
         'query',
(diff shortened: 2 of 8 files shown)


projects/client/src/lib/components/buttons/popup/_internal/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const POPUP_Z_INDEX = 777;
```
vladjerca: Why don't we have this as an actual css variable?

I was actually thinking of creating a `layers.css` file in which we extract all `z-index` values in order to reference them later on.

So my question is why do we have it as a JS constant and later bind it as a css var? 😕

vladjerca (changes requested):

vladjerca (approved):

vladjerca (approved):

PR #421, opened 2025-02-09
Feat/popular lists

seferturan:
## 🎶 Notes 🎶

- Adds a popular lists section in the show and movie summary pages.
- Viewing the lists will be a follow up PR.
- Loading/empty state also needs to be followed up:
  - Performance was not good enough to fetch the lists while requesting the rest of the summary data.
  - For now placed at the bottom to reduce the annoyance of possible content jumps.

## 👀 Examples 👀
Screenshot 2025-02-10 at 00 34 35

Screenshot 2025-02-10 at 00 35 06

### DIFF
--- a/projects/client/src/lib/sections/summary/components/lists/Lists.svelte
+++ b/projects/client/src/lib/sections/summary/components/lists/Lists.svelte
@@ -1,0 +1,74 @@
+<script lang="ts">
+  import SectionList from "$lib/components/lists/section-list/SectionList.svelte";
+  import * as m from "$lib/features/i18n/messages.ts";
+  import RenderFor from "$lib/guards/RenderFor.svelte";
+  import type { MediaType } from "$lib/requests/models/MediaType";
+  import ListSummaryCard from "./_internal/ListSummaryCard.svelte";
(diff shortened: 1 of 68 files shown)


projects/client/src/lib/sections/summary/components/lists/Lists.svelte
```diff
@@ -0,0 +1,74 @@
...
+  const MAX_MOBILE_LISTS = 3;
+
+  const {
+    slug,
+    type,
+    title,
+  }: { slug: string; type: MediaType; title: string } = $props();
+
+  // Due to slow performance, we fetch the lists here instead of useMovie/useShow
+  const { isLoading, lists } = useListSummary({ slug, type });
```
vladjerca, line 18: Makes sense, we should probably lazy load anything that isn't directly related to the media items.

eg: related, lists, etc.

seferturan: Yep, especially since the lists reserve heights :D I'll make a note of it

vladjerca (approved):

PR #423, opened 2025-02-10
Chore/remove favorites

seferturan:


commitlint.config.js
```diff
@@ -43,7 +43,6 @@ module.exports = {
         'dropdown',
         'e2e',
         'episode',
-        'favorite',
```
vladjerca: @seferturan Not from the API definition though :trollface: 

Someone got overexcited 😂

vladjerca (changes requested):

commitlint.config.js
seferturan: image

PR #442, opened 2025-02-14
Feat / clickable directors & writers

seferturan:
## 🎶 Notes 🎶

- Directors & writers on summary pages are now clickable.
- People summary pages now contain a dropdown per list to view different positions they worked at (e.g. `Directing`, `Writing`, `Acting`, etc.
- By default, the credits list will be for the position they're known for.
- Other changes:
  - `SectionList.svelte` can now have `dynamicActions`. These are actions with a dynamic size and will be placed before the scroll buttons to keep the scroll buttons at a stable location.
  - Switched the order of the actions on the season list.
- For now I left some small utils in `CreditsList.svelte`: I want to change the whole thing in a later PR -> uplift the requests to the `PeopleSummary` component, move the position dropdown below the poster or something, and add it as a search param so specific views can be shared.

## 👀 Examples 👀
Screenshot 2025-02-14 at 21 03 08

Screenshot 2025-02-14 at 21 03 18

Screenshot 2025-02-14 at 21 03 49

Before / after:
Screenshot 2025-02-14 at 22 32 29

Screenshot 2025-02-14 at 22 32 03

### DIFF
--- a/projects/client/src/lib/components/lists/section-list/SectionList.svelte
+++ b/projects/client/src/lib/components/lists/section-list/SectionList.svelte
@@ -19,6 +19,7 @@
     title,
     item,
     empty,
+    dynamicActions,
     actions: externalActions,
   }: SectionListProps<T> = $props();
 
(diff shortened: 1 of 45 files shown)


projects/client/src/lib/components/lists/section-list/SectionList.svelte
```diff
@@ -19,6 +19,7 @@
     title,
     item,
     empty,
+    dynamicActions,
```
vladjerca, line 22: I'll leave it as is for now but ideally we have only an action input and somehow extend rather than having 2 different inputs.

Make a Todo for this item so we revisit later.

vladjerca (approved):

PR #443, opened 2025-02-15
fix: scrollbars in safari

seferturan:
## 🎶 Notes 🎶
- Safari (╯°□°)╯︵ ┻━┻
- Scrollbars never became visible.
- Dropdowns were hidden under the section lists.

## 👀 Example 👀
I removed the `opacity:0`, which causes safari to be buggy when it comes to displaying hiding scrollbars. Example:
https://github.com/user-attachments/assets/4929838c-f989-484d-9904-095a4e9e7cf3

So for now, the overflow-x on the lists is set on a hover.

Dropdown fix:
Before/after:
Screenshot 2025-02-15 at 16 27 22

Screenshot 2025-02-15 at 16 27 31

### DIFF
--- a/projects/client/src/lib/components/dropdown/DropdownList.svelte
+++ b/projects/client/src/lib/components/dropdown/DropdownList.svelte
@@ -111,6 +111,7 @@
       }
 
       .trakt-list {
+        z-index: var(--layer-menu);
         opacity: 1;
       }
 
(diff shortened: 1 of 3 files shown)


seferturan:
Drafted; when on hover setting the overflow-x still hides the dropdown

vladjerca (approved):

PR #453, opened 2025-02-17
Feat/switch toggle

seferturan:
## 🎶 Notes 🎶

- Adds a switch toggle component
- Also added it to the design systems: http://localhost:5173/_design_system/buttons/toggles
- I played around a bit with an intl provider, but even with short `on`/`off` labels, certain translations will result in long values. We can revisit later, for now I opted to add an `innerText` property.

## 👀 Examples 👀
Screenshot 2025-02-17 at 20 11 50

https://github.com/user-attachments/assets/f8fb3002-fcdb-484b-ae2d-f03482f2d97d

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -70,6 +70,7 @@ module.exports = {
         'social',
         'spoiler',
         'summary',
+        'toggle',
         'trending',
         'up-next',
         'upcoming',
--- a/projects/client/src/app.d.ts
+++ b/projects/client/src/app.d.ts
@@ -50,6 +50,10 @@ declare global {
     onerror?: (event: Event) => void;
   };
 
+  type HTMLInputElementProps = {
+    value?: string;
+  };
+
   type HTMLElementProps = {
     disabled?: boolean;
     'data-testid'?: string;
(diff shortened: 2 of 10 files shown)


projects/client/src/lib/components/buttons/Switch.svelte
```diff
@@ -0,0 +1,207 @@
...
+
+<button
+  class="trakt-switch"
+  role="switch"
+  aria-checked={$isChecked}
+  aria-label={label}
+  onclick={() => isChecked.update((value) => !value)}
+  data-color={color}
+  {...props}
+>
```
vladjerca: Typically toggles are defined on top of `input[type=checkbox]` rather than over a `button` 👀

vladjerca (changes requested):

projects/client/src/lib/components/buttons/Switch.svelte
seferturan: Done 👌

seferturan:
Some more changes:

- It is no longer in `buttons`. Design system route has been adjusted accordingly.
- More semantically correct HTML.
- Hover animation is now temporary; and more a nudge/preview of the direction it will go in.

vladjerca (approved):

PR #504, opened 2025-02-26
fix(rating): ignore vote count if there is no rating

seferturan:
## ♩ Note ♩

- Apparently we can have empty ratings, but still have vote counts.

## 👀 Example 👀
Before:
Screenshot 2025-02-26 at 21 33 46

After:
Screenshot 2025-02-26 at 21 32 37

### DIFF
--- a/projects/client/src/lib/requests/_internal/mapToMediaRating.ts
+++ b/projects/client/src/lib/requests/_internal/mapToMediaRating.ts
@@ -19,11 +19,6 @@ export function mapToMediaRating(
       votes: ratings.trakt.votes,
       distribution: ratings.trakt.distribution,
     },
-    tmdb: {
-      rating: (ratings.tmdb?.rating ?? 0) / 10,
-      votes: ratings.tmdb?.votes ?? 0,
-      url: forceHttps(ratings.tmdb?.link),
-    },
     rotten: {
       critic: ratings.rotten_tomatoes?.rating ?? 0,
       audience: ratings.rotten_tomatoes?.user_rating ?? 0,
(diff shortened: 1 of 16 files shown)


projects/client/src/lib/requests/_internal/mapToMediaRating.ts
```diff
@@ -13,6 +13,17 @@ function forceHttps(url: string | Nil): HttpsUrl | Nil {
 export function mapToMediaRating(
   ratings: RatingResponse,
 ): MediaRating {
+  const mapToVoteCount = (
+    votes: number | Nil,
+    rating: number | Nil,
+  ): number => {
+    if (!rating || !votes) {
```
vladjerca: Is this correct though? 

aka: can you have something rated as 0 on IMDb?

If it's a data problem I would rather fix that 👀

vladjerca: Yeah, you see. On IMDb not only does it have a rating:

https://m.imdb.com/title/tt11162260/

But that 77 count is actually the number of lists the movie is part of @seferruran

seferturan: 78 votes ;) The only 77 is I see is the rise in popularity rank. If I click on user lists (https://m.imdb.com/lists/tt11162260/?ref_=tt_urls_sm) there are 267 lists.

This wasn't a case of being rated as 0, but we get a response where the rating is `null`, while still having a vote count. But even if it's outdated data, it's still weird to have a vote count, and not a rating.

@rectifyer, where do we get our imdb rating data from and how often is it updated? Does the vote count actually represent the number of ratings? It is kinda strange that this movie had a count for a `null` rating :P

seferturan:
Changed a bit; ratings are only mapped if they are deemed valid ratings.
Also removed some unused code.

projects/client/src/lib/requests/_internal/mapToMediaRating.ts
seferturan: @rectifyer, nvm, omdb returns it like that. Maybe something to also account for on OG when parsing the `n/a` response?

vladjerca (approved):

PR #506, opened 2025-02-27
Fix/list scroll pos

seferturan:
## 🎶 Notes 🎶

- Changes some list id's to be more verbose.
- Fixes the scroll history provider and action to always restore all lists on a page using correct id's.
  - Left a FIXME in there; ideally we get rid of the debounce, but would probably need a different approach. Maybe something like: we keep track of restores for a page; if a list on that page is being mounted/updated, it checks if that page had a restore 🤔 

## 👀 Example 👀

https://github.com/user-attachments/assets/c6aeee43-af0f-44b1-9de0-141b6554c26d


vladjerca (approved):
I'll accept this current fix for the time being. @seferturan I encourage you to revisit the implementation if needed with a different approach.

Ideally we should have this system tested as it's a nasty one if we don't automatically account for mounts and updates. Testing manually is a B.

PR #510, opened 2025-02-28
refactor / extract pagination query list store and make pagination params mandatory

vladjerca:
NOTE: also updated the default page size to 25, seems like a better chunk to serve.

Future: Default page size could be adaptive based on consumer (Eg; 25 on desktop, 15 tablet, 5 on mobile). Although this is a bit overkill so not tackling it now.

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -67,6 +67,7 @@ module.exports = {
         'query',
         'rating',
         'readme',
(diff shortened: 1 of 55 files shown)


seferturan (approved):
I applied a small fixup to also use `usePaginatedListQuery` in `useListItems`.

PR #521, opened 2025-03-03
Feat/more analytics

seferturan:
## 🎶 Notes 🎶

- Adds analytics events when changing the thema and locale picker.
- Adds analytics events when using media actions.
- Adds a rudimentary mapping from event types to data types.


seferturan:
Drafted; gonna revisit that `void`.

vladjerca (approved):

PR #522, opened 2025-03-04
fix(footer): allow copyright text to grow

vladjerca:
Before:
image

After:
image


seferturan (approved):
Applied a small fixup to keep the theme/locale picker on mobile/small tablet.

PR #533, opened 2025-03-13
Feat/underlined buttons

seferturan:
## ⚠️ Draft ⚠️
@anodpixels:

- With certain page backgrounds, the non active button is slightly less readable since the button itself no longer has a background. See examples below. Is this good enough to start with?
- By default, only buttons that can have an active state will show the colored underline. Non active, and regular buttons will fall back to the slightly grayer font color. The color will be shown on a hover. See video example below. These do seem a bit like disable ghost buttons now though. Not sure if we should fix this right this moment, since we only have use for the ones that can have an active state.
- You can check out this branch if you want to play around with it first.

## 🎶 Notes 🎶

- Adds an underlined style to the button component.
- Works best when used with links.

## 👀 Examples 👀
Screenshot 2025-03-13 at 15 30 42

Screenshot 2025-03-13 at 15 30 46

Screenshot 2025-03-13 at 15 30 51

https://github.com/user-attachments/assets/1b8a2c67-22ba-4a47-9f9b-13efa58a5f47

### DIFF
--- a/projects/client/src/lib/components/buttons/Button.svelte
+++ b/projects/client/src/lib/components/buttons/Button.svelte
@@ -44,7 +44,8 @@
   <div class="button-label">
     <p
       class="ellipsis"
-      class:small={subtitle != null}
+      class:small={subtitle != null || style === "underlined"}
+      class:bold={style === "underlined"}
       class:capitalize={text === "capitalize"}
     >
       {@render children()}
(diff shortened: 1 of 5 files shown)


seferturan:
After the meeting: we decided to have the labels always be the foreground color to increase readability.

PR #539, opened 2025-03-17
Fix/no auth queries

seferturan:
## 🎶 Notes 🎶

- ~Do not request user data if we know the user is not signed in~
- Do not show the watched history list if we know the user is not signed in

## 👀 Examples 👀

**~Before/after~**:
Screenshot 2025-03-17 at 19 00 56

Screenshot 2025-03-17 at 19 05 13

**Before**:
Screenshot 2025-03-17 at 19 08 37

And after 3 failed requests it would show:
Screenshot 2025-03-17 at 19 08 45

**After**:
Screenshot 2025-03-17 at 19 08 16


seferturan:
(╯°□°)╯︵ ┻━┻

Drafted, need to investigate the tests a bit more

seferturan:
Omitting the `useUser` changes for now.

PR #571, opened 2025-04-10
Feat/list action

seferturan:
## 🎶 Notes 🎶

- Makes it possible for users to add movies/shows to personal lists (on summary pages).
  - For now it's an internal component. When/if we ever have use of this in other places/forms, we can extract it and make it a media-action component.
- The dropdown is only shown if a user has lists, otherwise the `add to watchlist` button is shown.
- Follows the existing layout, i.e. will be placed at where the `add to watchlist` button is.
- Deviates a bit from the design (cc @anodpixels)
  - I followed youtube a bit more; i.e. not adding to watchlist by default. It is really strange that it always toggles the watchlisted state if I just want to add it to a list.
  - We can do a follow-up here and change it later. Maybe a separate action after all? Dialog with checkboxes? etc/

## 👀 Examples 👀

https://github.com/user-attachments/assets/48c43cec-008f-4e57-b97e-dc4b7632602f

Screenshot 2025-04-10 at 10 43 14

When there are no lists:
Screenshot 2025-04-10 at 10 36 49

### DIFF
--- a/projects/client/src/lib/sections/summary/components/media/MediaSummary.svelte
+++ b/projects/client/src/lib/sections/summary/components/media/MediaSummary.svelte
@@ -20,6 +20,9 @@
   import type { Snippet } from "svelte";
   import MediaDetails from "../details/MediaDetails.svelte";
   import MediaStreamingServices from "../details/MediaStreamingServices.svelte";
+  import ListDropdown from "../list-dropdown/ListDropdown.svelte";
+  import type { ListDropdownProps } from "../list-dropdown/ListDropdownProps";
+  import { useAllPersonalLists } from "../list-dropdown/useAllPersonalLists";
   import MediaMetaInfo from "../media/MediaMetaInfo.svelte";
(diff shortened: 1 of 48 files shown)


projects/client/src/lib/sections/summary/components/media/MediaSummary.svelte
```diff
@@ -54,27 +57,35 @@
...
 
 {#snippet mediaActions(device: "mobile" | "other" = "other")}
-  <WatchlistAction
-    {...watchlistProps}
-    style={device === "mobile" ? "action" : "normal"}
-  />
+  {#if $lists.length === 0}
+    <WatchlistAction
+      {...watchlistProps}
+      style={device === "mobile" ? "action" : "normal"}
```
vladjerca: Can we calculate the style top level and use everywhere?

vladjerca (changes requested):

vladjerca (approved):

PR #573, opened 2025-04-10
feat / light theme part #1

vladjerca:
- shadows for card covers
- shadows for list
- cover image and logo cover are now grayscale to blend into the background better
- card background

TODO (part #2):

- chips / tags
- secondary buttons


seferturan:
`cover image and logo cover are now grayscale to blend into the background better` Love this, looks great :D

PR #668, opened 2025-05-02
feat(list): add support for collapse expand

vladjerca:
https://github.com/user-attachments/assets/2ce41edd-a03b-49c4-891b-4ebf4455a714

https://github.com/user-attachments/assets/f26ca7fc-31eb-4659-bca2-f0c25da3c352


seferturan (approved):
Applied a small fixup to not use the container height as the list height. Fixes this issue:

https://github.com/user-attachments/assets/addad2f1-612a-456f-a633-5ea0c8930e2b

PR #677, opened 2025-05-06
feat(search): split results into movies and shows

vladjerca:
Follow-ups (most likely next week):

- Add People support
- Add `View All` per category


seferturan (approved):
Applied a fixup for the test 👌 

Not 100% happy about the list titles, but we can fine tune later. Maybe dedicated translation keys so we can properly translate `search results for movies`, or `Movie search results` 🤔

PR #685, opened 2025-05-07
fat / add season list to show view

vladjerca:
Implements a season list on the show details view, allowing users to select and browse episodes within a specific season.

- Introduces a new `SeasonList` component to display seasons with their posters.
- Includes a `SeasonEpisodeList` to show episodes for the selected season.
- Updates the show view to display the season list.
- Adds an endpoint to fetch seasons with images.


seferturan (dismissed):
Applied a fixup to add the translations 👌

PR #691, opened 2025-05-08
fix(query): accept 204 also as a valid response status

seferturan:


vladjerca (approved):

projects/client/src/lib/features/query/_internal/isSuccessResponse.ts
```diff
@@ -3,10 +3,12 @@ import type {
...
+const SUCCESSFUL_STATUSES: number[] = [200, 204] as const;
+
 export function isSuccessResponse<TInput>(
   response: RequestResponse<TInput>,
 ): response is SuccessResponse<TInput> {
   return Array.isArray(response)
-    ? response.every((item) => item?.status === 200)
-    : response.status === 200;
+    ? response.every((item) => SUCCESSFUL_STATUSES.includes(item?.status))
+    : SUCCESSFUL_STATUSES.includes(response.status);
```
vladjerca, line 13: Actually in case of 204 we should be returning null not mapping the data 😞

PR #715, opened 2025-05-14
Fix/side navbar nav

seferturan:
## 🎶 Notes 🎶

- Fixes some d-pad navigation issues:
  - It can now wait for dynamic content
  - On main pages, it will wait for the lists to be available, and select the first item instead of opening the side navbar.
  - Similarly for trakt pages, e.g. when going to a summary page. It will now select the first action button instead.
- If it times out, it will fall back to it's old behavior and open the sidebar.

## 👀 Examples 👀
Before:

https://github.com/user-attachments/assets/37757e84-7a02-4443-98bc-9a5258d508e6

After:

https://github.com/user-attachments/assets/2f0fcf98-e7c9-4bb2-98b6-4453cbe1867e

### DIFF
--- a/projects/client/src/lib/components/lists/section-list/ShadowList.svelte
+++ b/projects/client/src/lib/components/lists/section-list/ShadowList.svelte
@@ -16,6 +16,8 @@
   import ExpandIcon from "./ExpandIcon.svelte";
   import { scrollTracking } from "./scrollTracking";
 
+  const EMPTY_STATE_CLASS = "shadow-list-empty-state";
+
   type SectionListProps<T> = ListProps<T> & {
     subtitle?: string;
     empty?: Snippet;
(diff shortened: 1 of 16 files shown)


projects/client/src/lib/components/lists/section-list/ShadowList.svelte
```diff
@@ -88,6 +90,7 @@
   class:shadow-list-container-collapsed={$isCollapsed}
   class:shadow-list-container-mounted={$isMounted}
   class:shadow-list-container-no-header={!isHeaderVisible}
+  data-dynamic-content={`[data-dpad-navigation="${DpadNavigationType.Item}"], .${EMPTY_STATE_CLASS}`}
```
vladjerca: Let's call this `data-dynamic-selector`

vladjerca (dismissed):

vladjerca (approved):

PR #717, opened 2025-05-14
fix / disable paraglide.js telemetry in CI

vladjerca:
Disables PostHog telemetry within Paraglide.js during CI builds.

- Patches the Paraglide.js library to prevent telemetry data from being sent during CI runs. This avoids potential issues with telemetry interfering with the build process and causing timeouts.
- Configures Deno to use a local cache to speed up installations and ensure reproducibility.


vladjerca:
@seferturan works now:

image

PR #718, opened 2025-05-14
fix(streaming): map to more usable netflix and disney links

seferturan:
## 🎶 Notes 🎶

- Better links for netflix and disney.
- Movies go to the movie page, episodes unfortunately take you to the show page (but still an improvement over not opening at all or taking you to the home dashboard).
- For netflix to work, it depends on: https://github.com/trakt/trakt-android/pull/101

### DIFF
--- a/projects/client/src/lib/requests/_internal/mapToDirectLink.ts
+++ b/projects/client/src/lib/requests/_internal/mapToDirectLink.ts
@@ -1,0 +1,24 @@
+import type { WatchNowServiceResponse } from '@trakt/api';
+
+export function mapToDirectLink(response: WatchNowServiceResponse) {
+  const { source, link_direct } = response;
+
+  if (!link_direct) {
+    return;
+  }
+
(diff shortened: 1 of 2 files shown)


projects/client/src/lib/requests/_internal/mapToDirectLink.ts
```diff
@@ -0,0 +1,24 @@
+import type { WatchNowServiceResponse } from '@trakt/api';
+
+export function mapToDirectLink(response: WatchNowServiceResponse) {
```
vladjerca, line 3: I hope we move this mapping serverside @seferturan

vladjerca (approved):

projects/client/src/lib/requests/_internal/mapToDirectLink.ts
seferturan: Yep, once everything is working good enough, this logic can be moved. The links sent from the server should be correct ones 👌

PR #729, opened 2025-05-20
Refactor/summary queries

seferturan:
## 🎶 Notes 🎶

- Extracts meta info queries to the component.
- ~~This is pt1; still wanna do a follow-up for the episode summary page to fetch the show data in parallel.~~
- Parallelizes queries used for summary pages
- Could use some further refactoring to reduce some more duplication.

## 👀 Examples 👀
Before/after:
before_episode

after_episode

Before/after:
before_show

after_show

Before/after:
before_movie

after_movie


seferturan:
Woops, need to update the tests :P Drafted until then

PR #753, opened 2025-06-04
chore / updates dependencies to latest versions

vladjerca:
This PR updates both development and production dependencies to their latest minor and major versions.

This ensures the project benefits from the latest features, bug fixes, and security improvements.

msw@2.9.0 throws, so keep it at 2.7.3 for the time being


seferturan (approved):
Applied some fixups:

- In case of `@sveltejs/kit`, I change the update to 2.21.0 instead of the latest:
  - The latest had this issue (https://github.com/sveltejs/kit/issues/13806). Which is fixed in the newer deno version. But the latest deno version has that `markResourceTiming` issue. Which is fixed, but not yet released. Once that's released, we can upgrade both.
- The tooltip had a new interface, I updated that also.
- There is still a typing issue, I left a fixme in `mapToMediaVideo.ts`. I didn't want to block this PR since it only was a typing issue.

Issue #755, opened 2025-06-04
Watchlist only shows first 500 titles


dzg:
### Code of Conduct

- [x] I have read and agree to the project's [Code of Conduct](https://github.com/trakt/trakt-lite/blob/main/.github/CODE_OF_CONDUCT.md)

### Where did you encounter the bug?

https://app.trakt.tv/users/me/lists/movies?status=all

### Please describe the bug below

Viewing "all" in watchlist movies only shows the first 500 titles. And no pagination to load more.

vladjerca:
@dzg we'll be looking at lazy loading, we know there's a limitation with this ATM.

Will keep this thread updated when we address this 👍

PR #784, opened 2025-06-13
Feat/lazy load

seferturan:
## 🎶 Notes 🎶

- Add support for lazy loading `DrilledMediaList`.
  - If we ever need this for horizontal lists, extracting it and adding horizontal support should be fairly simple.
  - Accounts for resizes also, and will keep loading until the screen is full enough.
  - Shows a small loading indicator when loading new data.
- Added a temporary fix to account for broken `trending` data. Will be removed when server is fixed.

## 👀 Examples 👀

https://github.com/user-attachments/assets/ff496ca1-338f-46a9-ba14-f5b5863749e2

https://github.com/user-attachments/assets/5b796c8e-bb1c-428c-b2c0-c054b6d7e49d


vladjerca (approved):

vladjerca:
Fixes: #755

PR #880, opened 2025-07-24
fix(up-next): dealing with missing episode runtime

seferturan:
## 🎶 Notes 🎶

- Forgot this after the schema changes 😅
- Small fix to the progress tag to not wrap durations

### DIFF
--- a/projects/client/src/lib/sections/lists/components/EpisodeItem.svelte
+++ b/projects/client/src/lib/sections/lists/components/EpisodeItem.svelte
@@ -49,7 +49,7 @@
           <span class="ellipsis">
             {EpisodeIntlProvider.remainingText(props.episode.remaining)}
           </span>
-          <span>
+          <span class="no-wrap">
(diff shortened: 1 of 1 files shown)


projects/client/src/lib/requests/queries/sync/upNextQuery.ts
```diff
@@ -54,7 +54,7 @@ const upNextRequest = (params: UpNextParams) => {
 export function mapUpNextResponse(item: UpNextResponse[0]): UpNextEntry {
   const show = mapToShowEntry(item.show);
   const episode = mapToEpisodeEntry(item.progress.next_episode);
-  episode.runtime ??= show.runtime;
+  episode.runtime = isNaN(episode.runtime) ? show.runtime : episode.runtime;
```
vladjerca, line 57: @seferturan I think we should just default to the show runtime here in the mapper.

vladjerca (changes requested):

PR #891, opened 2025-07-30
Feat/comment reactions

seferturan:
## 🎶 Notes 🎶

- Adds support for opening popup menus in dialogs.
  - Since dialogs live in the top layer, the portal needs to add things to the dialog instead of the body.
- Adds v0 for comment reactions.
  - Design is a mix of what was in figma + some linkedin inspirations.
  - Single reaction allowed.
  - Follow-ups we should do: animated emojis, reaction counts, adding support to the portal to set a popup anchor.
- Removes the comment likes code.

## 👀 Examples 👀

https://github.com/user-attachments/assets/4ffe714a-0a85-4592-83c9-4e6c1c8c9025


projects/client/src/lib/features/portal/_internal/usePopupHelpers.ts
```diff
@@ -47,9 +47,13 @@ export function usePopupHelpers() {
   };
 
   const addHelpers = (target: HTMLElement) => {
+    const openDialog = document.querySelector('dialog[open]');
```
vladjerca: Why not

const target = queySelector... ?? body

PR #893, opened 2025-07-30
feat(i18n): exclude android / ios from all keys and make en.json sso

vladjerca:
Updates the i18n generation process to support platform-specific exclusions defined in the `en.json` file.

- Modifies the generator to augment non-English locale files with platform configuration from `en.json`, allowing for selective inclusion/exclusion of messages based on the target platform.
- Excludes `android` and `ios` platforms from a large number of the default `en.json` keys.
- Enforces that `en.json` is used as the SSO.

This ensures that platform-specific logic is centralized in the `en.json` file, improving maintainability and consistency across locales.


seferturan (approved):
I applied a fixup for the failing generations & tests 👌

michaldrabik (approved):
🥇

PR #897, opened 2025-07-30
feat(list): adds tags to other watchlist based lists

seferturan:
## ♪ Note ♪

- All watchlist based lists are now tagged. For the released/unreleased ones, only a read-only tag for informational purposes.

## 👀 Example 👀
Screenshot 2025-07-30 at 21 33 00


vladjerca:
@seferturan check the `fixup!` 

image

PR #904, opened 2025-07-31
fix(i18n): use episode number label in all card variations

vladjerca:
Consistently applies the episode number formatting across various card components and sections.

- Introduces `episodeNumberLabel` utility function to encapsulate the formatting logic, ensuring consistency and easier maintenance.
- Replaces hardcoded formatting in `EpisodeCard.svelte` and `MediaSummaryCard.svelte` with the new utility.
- Moves `episodeActivityTitle` to the `intl` folder.


projects/client/src/lib/sections/lists/components/MediaSummaryCard.svelte
```diff
@@ -21,13 +22,22 @@
...
+  function footerTag() {
+    if (rest.type === "episode") {
+      return tag;
+    }
+
+    return undefined;
+  }
 </script>
 
+<!--
```
seferturan: Was this here for testing purposes, or can it be removed?

vladjerca: This can be removed. The tag was defined even when empty so moved logic to a method.

LE: Pushed.

PR #905, opened 2025-07-31
Feat/feature flags

seferturan:
## 🎶 Notes 🎶

- Adds support for using feature flags.
  - 1 flag at the moment, for reactions.
  - Adding a feature flag: add an entry to the `FeatureFlag` enum, and use the `RenderForFeature` guard with that flag.
- Reverts the removal of the `like comment` code.
- Places the comment reactions under the feature flag.
  - By default likes are used, reactions are shown when the flag is enabled.
- The button is placed in the footer. Having a fixed position button in the UI quickly got annoying.

## 👀 Examples 👀
Screenshot 2025-07-31 at 15 53 36

Screenshot 2025-07-31 at 15 54 09

https://github.com/user-attachments/assets/0371b81d-913d-4b0b-981f-efa4a06c4efe

### DIFF
--- a/projects/client/src/lib/features/analytics/events/AnalyticsEvent.ts
+++ b/projects/client/src/lib/features/analytics/events/AnalyticsEvent.ts
@@ -25,6 +25,7 @@ export const AnalyticsEvent = {
   Watchlist: buildEventKey(MEDIA_ACTION_PREFIX, 'watchlist'),
   List: buildEventKey(MEDIA_ACTION_PREFIX, 'list'),
   RemoveFromHistory: 'remove-from-history',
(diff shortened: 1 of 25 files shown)


projects/client/src/lib/features/feature-flag/useFeatureFlags.ts
```diff
@@ -0,0 +1,25 @@
...
+      );
+      return updatedFlags;
+    });
+  };
+
+  return {
+    flags: derived(flags, ($flags) => $flags),
+    setFlag,
+  };
+}
```
vladjerca: This should be `useFeatureFlag` not Flag`s` we should request a `specific` feature flag and have that reactive not the entire collection...

vladjerca (changes requested):

PR #910, opened 2025-08-04
Feat/plex from collection

seferturan:
## 🎶 Notes 🎶

- Adds the plex button on summary pages if an item is in the users plex collection.
- Feature flagged for now:
  - Currently placed in the same location as the previous iteration.
  - Not very usable on desktop; the link on desktop only takes you to the plex streaming service page :/

## 👀 Example 👀
Screenshot 2025-08-04 at 09 25 30


projects/client/src/lib/sections/summary/components/media/MediaMetaInfo.svelte
```diff
@@ -59,13 +60,15 @@
       audience="all"
       navigation="default"
     >
+      <PlayOnPlexButton {...target} style="normal" />
```
vladjerca: We should just default to plex if it's in your collection. Replace stream on

seferturan: Done 👌

PR #911, opened 2025-08-04
Feat/webos

seferturan:
## 🎶 Notes 🎶

- v0 for webOS (for webOS 24 and higher).
- Since webOS24 is based on Chromium 108, not every feature we use is available:
  - CSS nesting: added `postcss-nested` to deal with this.
  - Missing JS features: used `core-js` for polyfills (only when device is tv).
  - `color-mix`:  extracted almost all color-mixes, and switched to a fallback if color-mix is not available.
- Also contains some minor fixes of navigation issues I ran into while testing.
- First couple of follow-ups we should tackle afterwards:
  - Deal with alert/prompts (alert works in the simulator but not on the TV. prompts are blocked in both).
  - webOS 22 & 23
  - Release as closed beta (if possible).
  - UI scaling (seems as if it's rendered at 1080p and upscale to 4k).

## 🧪 Testing / Examples 👀
If you want to test it locally:

- Run a local dev environment.
- In the `index.html`, change the href to your localhost dev environment.
- Download and use [the webOS24 simulator](https://webostv.developer.lge.com/develop/tools/simulator-installation)
- In the simulator, choose `File` -> `Launch App`, and choose the `projects/web-os` directory. See below for an example.

https://github.com/user-attachments/assets/fa64ac5a-0834-4408-9f87-694d68a4191b

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,18 @@ adb reverse tcp:5173 tcp:5173
 
 **Coming Soon!**
 
+### webOS Development
+
+- Install the developer mode app on your TV, see the
+  [documentation](https://webostv.developer.lge.com/develop/getting-started/developer-mode-app)
+  - Alternatively, use the
+    [simulator app](https://webostv.developer.lge.com/develop/tools/simulator-installation)
+- Install the
+  [webOS CLI](https://webostv.developer.lge.com/develop/tools/cli-installation).
+- Refer to the
+  [CLI dev guide](https://webostv.developer.lge.com/develop/tools/cli-dev-guide)
+  to see how to create an app, launch it on your TV, etc.
+
 ## Build Trakt Web
 
 To build the Trakt Web client, run:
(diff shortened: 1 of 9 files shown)


projects/client/src/lib/features/filters/_internal/constants.ts
```diff
@@ -74,5 +74,5 @@ export const FILTERS = [
 ] as const;
 
 export const DEFAULT_TV_FILTERS = {
-  [FilterKey.Streaming]: 'any_subscriptions',
+  [FilterKey.Streaming]: 'any',
```
vladjerca: Why was this changed? `any` isn't the correct value.

seferturan: Any is the correct value: https://github.com/trakt/trakt-api/pull/551/files

seferturan: Oh, wait, I guess it should be `subscriptions` in this case.

vladjerca (approved):

PR #912, opened 2025-08-05
fix(summary): display TBA for unknown dates

vladjerca:
Refactors the AirDate tag to rely solely on the `airDate` property and introduces a utility function to determine if a date represents an unknown or TBA state. This removes the need for a separate `year` property and centralizes the logic for displaying "TBA" status.

- Uses `isMaxDate` to check if the airDate represents a "To Be Announced" state.
- Simplifies components using the AirDate tag by removing the year prop.


seferturan (approved):
Applied a small fixup for the import path 👌

PR #923, opened 2025-08-08
fix(calendar): force loading indicator desctruction

seferturan:
## ♪ Note ♪

- The loading indicator was the issue..

## 👀 Example 👀

Before:

https://github.com/user-attachments/assets/61e1f6af-9c86-41e4-9d16-9f090b1bed3a

After:

https://github.com/user-attachments/assets/d27f7500-0ac8-4a73-afdf-72994fe81200

### DIFF
--- a/projects/client/src/lib/features/calendar/Calendar.svelte
+++ b/projects/client/src/lib/features/calendar/Calendar.svelte
@@ -24,7 +24,14 @@
   </div>
 
   {#if $isLoading}
-    <LoadingIndicator />
+    <!--
+    Loading indicator is keyed to ensure destruction, since
+    the calendar items depend on accurate element positions onMount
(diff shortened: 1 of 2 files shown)


projects/client/src/lib/features/calendar/Calendar.svelte
```diff
@@ -24,7 +24,13 @@
...
 
   {#if $isLoading}
-    <LoadingIndicator />
+    <!--
+    Loading indicator is keyed to ensure destruction, since
+    the calendar items depend on accurate element positions onMount
+   -->
+    {#key "calendar-loading"}
+      <LoadingIndicator />
+    {/key}
```
vladjerca, line 34: This still feels weird 😅

LE: the fact that we need to do this indicates some other underlying issue.

PR #929, opened 2025-08-12
Routes API requests to HD domain

vladjerca:
- Ensures API requests are routed to the "hd." subdomain instead of "apiz." 
- Adds a build-time constant for the HD target to avoid hardcoding
- Skips replacement for requests including "watchnow"

### DIFF
--- a/projects/client/src/app.d.ts
+++ b/projects/client/src/app.d.ts
@@ -15,6 +15,7 @@ declare global {
   const TRAKT_CLIENT_ID: string;
   const TRAKT_MODE: 'development' | 'production' | 'test';
   const TRAKT_TARGET_ENVIRONMENT: Environment;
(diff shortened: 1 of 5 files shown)


seferturan:
Shows seem to be missing aired episodes:
Screenshot 2025-08-12 at 14 53 25

Not yet sure why, but images seem to have issues also:

Slower to load, before/after:

https://github.com/user-attachments/assets/00812bdc-327c-4dd7-8e05-b1aaa4f5cf6b

https://github.com/user-attachments/assets/bab72df7-1824-4dfb-a711-457627076099

Can flicker:

https://github.com/user-attachments/assets/a4eab70f-c72f-4a1c-beab-c4026b5d7cd1

seferturan (changes requested):
See comment. Not yet sure what causes the image issues.

seferturan (approved):

PR #949, opened 2025-08-15
feat(search): coalesce media search results

seferturan:
## 🎶 Notes 🎶

- Combines media search results into a single list.
- Results are shown in a grid.
- For now 2 separate entries in the context:
  - So when going from `people` back to `media`, it will re-use the media type.
  - We should do a follow up where we store these things.

## 👀 Example 👀

https://github.com/user-attachments/assets/fdee3838-d430-4331-8fd8-c37cf007eb4c


projects/client/i18n/meta/en.json
```diff
@@ -3354,6 +3354,14 @@
...
       ]
     },
+    "list_title_search_results": {
+      "default": "Results",
+      "description": "Text shown when displaying search results.",
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
```
vladjerca: Can't we use `text_search_results_for` ?

Considering that we have the inputed text visible at all times I'm even tempted to remove the title all-together 🤔 

Yeah, let's go with that... no more title.

projects/client/src/routes/search/+page.svelte
```diff
@@ -64,50 +74,38 @@
...
+
+          {#snippet item(result)}
+            <DefaultMediaItem
+              type={result.type}
+              media={result}
+              style="cover"
+              tag={result.type === "show" ? showTag : undefined}
+            />
+          {/snippet}
+        </SectionList>
```
vladjerca: Let's turn this into a `GridList` we have a lot of unused real-estate (let's keep display as posters).

projects/client/i18n/meta/en.json
seferturan: We could, but I opted to use just `Results` because:

- The buttons already show the state.
- To save space on smaller screens, it felt cramped with `text_search_results_for` and the buttons.

vladjerca (approved):

projects/client/i18n/meta/en.json
vladjerca: @seferturan yeah but no using a title at wall would save even more space 😝

vladjerca: Something like this? 

(also incorporates my point from https://github.com/trakt/trakt-web/pull/949#issuecomment-3193668387)

image

vladjerca (changes requested):
Let's address the issues first

projects/client/i18n/meta/en.json
seferturan: Hmmm, I like how clean it is, but this would no longer separate `media` from `people. But I think we could deal with that 🤔

- Or we'd have to put everything in one grid and allow for a combination of `movies`, `shows`, and `people` in the same list.
- Or only allow one of them to be selected.
- Or also add the media toggle there also, example:
image

Thoughts?

vladjerca (approved):

PR #969, opened 2025-08-20
feat(i18n): Android strings sync

michaldrabik:
- enable required strings already present
- add some new ones to fill the gaps
- use ellipsis symbol instead of 3 dots


projects/client/i18n/meta/en.json
```diff
@@ -5101,6 +5055,208 @@
...
+      },
+      "exclude": [
+        "ios",
+        "web"
+      ]
+    },
+    "tag_text_people_eager": {
+      "default": "{count} eager",
+      "description": "Text for the tag that indicates how many people are eager to watch a movie or show.",
+      "variables": {
```
vladjerca, line 5251: @michaldrabik @anodpixels @seferturan  I thought we removed these labels.

vladjerca:
@michaldrabik the history is screwed on this branch: 

image

projects/client/i18n/meta/en.json
michaldrabik: I think UI/UX boyz were about to review these icons/badges before final change, so until that I kept the original labels in place. Web is excluded though and I can make another update later.

seferturan: Ah, I forgot to remove the labels.

@michaldrabik, we can align on some icons for now, and Alex can do a pass over it later.

michaldrabik: Let it be then. I'll just make a new cleaner PR in a sec.

PR #1018, opened 2025-09-09
chore(social): update social icon

seferturan:
## ♪ Note ♪

- We now have a social icon, so the crappy version I made can be replaced.

## 👀 Example 👀
Before:
Screenshot 2025-09-09 at 16 23 56

After:
Screenshot 2025-09-09 at 16 24 45


seferturan:
Hmm, even though they have the same size, visually this icon seems larger. Thoughts @anodpixels?

vladjerca (approved):

PR #1024, opened 2025-09-10
fix(card): deemphasis on safari

seferturan:
## 🎶 Notes 🎶

- Fix for cards not being deemphasized in safari.
- Not sure why it doesn't work in safari when applying it on the card. Didn't want to deep dive into finding the root cause, for now this also fixes it.


vladjerca (approved):
Safari is insane...

PR #1044, opened 2025-09-16
Fix/retry token

seferturan:
## 🎶 Notes 🎶

- Small refactor to extract the retry delay function.
- Also use a retry mechanism when posting token.


projects/client/src/lib/features/auth/stores/_internal/postToken.ts
```diff
@@ -0,0 +1,18 @@
...
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify({
+          token: value,
+          expiresAt,
+        }),
+      }),
+    { maxTry: 5, delay: ({ currentTry }) => retryDelay(currentTry - 1) },
+  );
+}
```
vladjerca: Extract our own `retry` method as we discussed and set config in there please.

vladjerca (changes requested):

projects/client/src/lib/features/auth/stores/_internal/postToken.ts
seferturan: We don't have our own? We use the retry of `QueryClient` itself. I extracted the retryDelay function to keep that the same.

vladjerca: @seferturan I mean extract the `retry` util for any promises that need retried, the fact that we use tanstack for basic fetching is a parallel aspect.

seferturan: Ooooh, mf, I totally misread that 🤦

seferturan: done 👌

PR #1045, opened 2025-09-16
feat: added text_cta_personal_lists string

michaldrabik:
Example usage: 
image


seferturan (approved):
I applied a fixup to also generate the translations.

PR #1095, opened 2025-10-06
perf: preconnect to media.trakt.tv

seferturan:


projects/client/src/worker/Domain.ts
```diff
@@ -8,5 +8,6 @@ export const Domain = {
   ],
   images: [
     'walter-r2.trakt.tv',
+    'media.trakt.tv',
```
vladjerca, line 12: @seferturan add a fixme to remove walter once all sources are using media.*

vladjerca (changes requested):

PR #1188, opened 2025-10-28
fix: use unique keys for media and episodes

seferturan:
## 🎶 Notes 🎶

- Since our id's aren't unique enough, and since we used the id's as keys, mixed lists could result in duplicate keys.
- All media and episode items now have a unique key (i.e. the id prefixed by their type). 
  - ~~Can do a refactor later to ensure all items we put in lists have keys, and completely get rid of using id's.~~
- Also added a refactor to make keys mandatory in lists 

## 👀 Example 👀

Before:
Screenshot 2025-10-28 at 10 45 08

After:
Screenshot 2025-10-28 at 10 50 30


projects/client/src/lib/components/lists/grid-list/GridList.svelte
```diff
@@ -1,4 +1,4 @@
-<script lang="ts" generics="T extends { id: unknown }">
+<script lang="ts" generics="T extends { id: unknown; key?: string }">
```
vladjerca: @seferturan why not standardize on key everywhere?

vladjerca (changes requested):

projects/client/src/lib/components/lists/grid-list/GridList.svelte
seferturan: To not block the fix on a bigger change, hence the `Can do a refactor later to ensure all items we put in lists have keys, and completely get rid of using id's.`.
Buuut, since this is in discover mode only, I guess it could be part of this also since it doesn't effect actual users.

seferturan: Aaaand done

PR #1332, opened 2025-11-28
feat(strings): general client strings sync #2

michaldrabik:
- another batch of strings sync with Android client to catch up
- traktify action run before PR

### DIFF
--- a/projects/client/i18n/meta/en.json
+++ b/projects/client/i18n/meta/en.json
@@ -2960,7 +2960,7 @@
       ]
     },
     "textarea_placeholder_comment": {
-      "default": "Review...",
+      "default": "Add a review...",
       "description": "Placeholder text for the review textarea when adding a new review.",
       "exclude": [
         "ios"
@@ -2998,7 +2998,7 @@
       ]
     },
     "translated_value_error_comment_invalid_content": {
-      "default": "Comments must be at least 5 words.",
+      "default": "Reviews must be at least 5 words.",
       "description": "Error message shown when a review is invalid due to content length.",
       "exclude": [
         "ios"
@@ -4202,7 +4202,7 @@
       ]
     },
     "text_about": {
-      "default": "About",
+      "default": "About me",
       "description": "Text for the section that allows users to change their 'about me' description.",
       "exclude": [
         "ios"
--- a/projects/client/i18n/meta/bg-bg.json
+++ b/projects/client/i18n/meta/bg-bg.json
@@ -1102,7 +1102,7 @@
       "default": "Добави ново ревю"
     },
     "textarea_placeholder_comment": {
-      "default": "Твоето ревю..."
+      "default": "Добавете рецензия..."
     },
     "switch_label_mark_as_spoiler": {
       "default": "Маркирай като спойлер"
@@ -1117,7 +1117,7 @@
       "default": "Публикувай отговор"
     },
     "translated_value_error_comment_invalid_content": {
-      "default": "Ревютата трябва да съдържат поне 5 думи."
+      "default": "Рецензиите трябва да са поне 5 думи."
     },
     "translated_value_error_reply_invalid_content": {
       "default": "Отговорите трябва да са поне 5 думи."
@@ -2321,9 +2321,6 @@
     },
     "button_label_view_all_lists": {
       "default": "Виж всички списъци"
-    },
-    "list_description_released_movies": {
-      "default": "Излезли филми"
     }
   }
 }
--- a/projects/client/i18n/meta/da-dk.json
+++ b/projects/client/i18n/meta/da-dk.json
@@ -1102,7 +1102,7 @@
       "default": "Tilføj en ny anmeldelse"
     },
     "textarea_placeholder_comment": {
-      "default": "Din anmeldelse..."
+      "default": "Tilføj en anmeldelse..."
     },
     "switch_label_mark_as_spoiler": {
       "default": "Marker som spoiler"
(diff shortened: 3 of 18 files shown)


projects/client/i18n/meta/en.json
```diff
@@ -2942,14 +2934,29 @@
         "ios"
       ]
     },
+    "dialog_title_comments_warning": {
+      "default": "Minimum 5 words, be respectful, mark spoilers.",
```
seferturan: Do we need to pre-warn users? 😅 For the errors, we have translated values (`translated_value_error_comment_invalid_content` & `translated_value_error_comment_unknown`)

projects/client/i18n/meta/en.json
```diff
@@ -2942,14 +2934,29 @@
...
     "dialog_title_comment": {
       "default": "Review",
       "description": "Title for the review dialog, which is used when a user is adding a new review.",
       "exclude": [
-        "android",
         "ios"
       ]
     },
+    "dialog_title_reply": {
+      "default": "Reply",
```
seferturan: Isn't reply supposed to be inline, instead of a dialog? 
Example:
Image

projects/client/i18n/meta/en.json
```diff
@@ -4083,6 +4086,14 @@
         "web"
       ]
     },
+    "button_text_submit": {
```
seferturan: This might also not be needed with icon only buttons.

projects/client/i18n/meta/en.json
```diff
@@ -4123,6 +4134,14 @@
         "web"
       ]
     },
+    "list_description_released_shows": {
+      "default": "Released shows",
```
seferturan: Didn't we get rid of these `released` lists? 🤔 Same for the `movies` and `shows & movies` ones.

projects/client/i18n/meta/en.json
```diff
@@ -5626,6 +5712,14 @@
         "ios"
       ]
     },
+    "list_title_from_watchlist": {
+      "default": "From Watchlist",
```
seferturan: Same here? Is this `Start watching`? 😅

projects/client/i18n/meta/en.json
```diff
@@ -5938,23 +6032,28 @@
       "default": "Watched",
       "description": "Text for the tag that is shown before the movie or show watch count. This should be as short as possible.",
       "exclude": [
-        "android",
         "ios"
       ]
     },
+    "tag_text_watched_count": {
+      "default": "Watched: {count}",
```
seferturan: Here we should probably align, I thought we had switched back to using the `dot` instead of `:`.

Atm, in v3 I have `tag_text_watched`, and append the dot and count separately; at the time was done like that so I could animate the number individually.

projects/client/i18n/meta/en.json
```diff
@@ -6038,6 +6133,22 @@
         "ios"
       ]
     },
+    "text_this_month_so_far": {
+      "default": "So far, this month",
```
seferturan: Magna wanted to get rid of `so far`, you can use `text_this_month`.

projects/client/i18n/meta/en.json
```diff
@@ -6038,6 +6133,22 @@
...
+      "description": "Text shown to indicate the users stats of the current month so far.",
+      "exclude": [
+        "web",
+        "ios"
+      ]
+    },
+    "text_this_month_in_review": {
+      "default": "{title} in review",
```
seferturan: `Month in review`?

projects/client/i18n/meta/en.json
michaldrabik: I can do dots.

projects/client/i18n/meta/en.json
michaldrabik: image

 Would be for this one - but I see the mistake in definition yeah, will be:
 
 ```
    "text_this_month_in_review": {
      "default": "{month} in review",
      "description": "Text shown to indicate the users stats of the given month review.",
      "variables": {
        "month": {
          "type": "string"
        }
      }
    },
 ```

projects/client/i18n/meta/en.json
michaldrabik: Thought why not as we know this is a rule and it would be one less friction and red error for user. Also got a nice subtitle space for it and submitting is only allowed if comment has enough words. Text itself is from og.

TBD on next sync imo:

image

projects/client/i18n/meta/en.json
michaldrabik: For the primary style button I will need something but I can use "Add review" which is already there in the bank.

image

projects/client/i18n/meta/en.json
michaldrabik: Was making this case of "Start Watching" not working great with movies during last sync but TBD on next one will group ask this again and revert if so. This would also inform user better where does the content of this section originates from.

image

vladjerca: The conclusion was to use `Start Watching`, there was no open ended conversation.

michaldrabik: alrighty

projects/client/i18n/meta/en.json
michaldrabik: It's been used as a sorting indicator of Start Watching list but if you think its not needed I'll scrap that too.

projects/client/i18n/meta/en.json
seferturan: Ah, there we'd switched to showing the month only:
image

PR #1337, opened 2025-12-01
feat(streaming): use colored svg logos for well known services

seferturan:
## 🎶 Notes 🎶

- Removes the `mask` hack for colored streaming service logo's in favor of using SVG logo's for well known services.
  - We can expand as we go on.

## 👀 Example 👀
Screenshot 2025-12-01 at 08 37 30


projects/client/src/lib/components/media/streaming-service/_internal/constants/index.ts
```diff
@@ -0,0 +1,22 @@
...
+  crunchyroll: ['crunchyroll'],
+  disneyPlus: ['disney_plus'],
+  googlePlayMovies: ['google_play_movies'],
+  hulu: ['hulu'],
+  netflix: ['netflix'],
+  paramountPlus: ['paramount_plus'],
+  peacock: ['peacock'],
+  plex: ['plex'],
+  rakutenTv: ['rakuten_tv'],
+  youtube: ['youtube'],
```
vladjerca, line 21: @seferturan shouldn't you then be using an includes check for these as well?

seferturan: I should indeed. Done 👌

PR #1354, opened 2025-12-03
chore(guards): guard dev mode users as a director

seferturan:
## ♪ Note ♪

- Updates the guard to see dev mode users as a director

### DIFF
--- a/projects/client/src/lib/guards/_internal/RenderForAudience.svelte
+++ b/projects/client/src/lib/guards/_internal/RenderForAudience.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
   import { useAuth } from "$lib/features/auth/stores/useAuth";
   import { useUser } from "$lib/features/auth/stores/useUser";
+  import { IS_DEV } from "$lib/utils/env";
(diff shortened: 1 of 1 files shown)


projects/client/src/lib/features/feature-flag/FeatureFlagTool.svelte
```diff
@@ -60,7 +61,15 @@
...
   {/if}
-</RenderFor>
+{/snippet}
+
+{#if IS_DEV}
+  {@render featureFlagToolButton()}
+{:else}
+  <RenderFor audience="director">
+    {@render featureFlagToolButton()}
+  </RenderFor>
```
vladjerca: @seferturan just a thought, but what if a `dev` is implicitly a `director`:

We could in the `RenderForAudience`:

```ts
      (audience === "director" && $isAuthorized && ($user?.isDirector || IS_DEV))
```

seferturan: I like that, done 👌

PR #1356, opened 2025-12-03
feat(button): ghost are always transparent

seferturan:
## 🎶 Notes 🎶

- Removes the blur from ghost buttons; they're always transparent now.
- ~~Kept the blur in summary cards since it can clash with some background cover images.~~

### DIFF
--- a/projects/client/src/lib/components/buttons/ActionButton.svelte
+++ b/projects/client/src/lib/components/buttons/ActionButton.svelte
@@ -166,9 +166,6 @@
     &[data-style="ghost"] {
       background-color: transparent;
 
-      /** This is required for improved readability when rendering over a cover image */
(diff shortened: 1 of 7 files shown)


projects/client/src/lib/sections/lists/components/_internal/SummaryCardBottomBar.svelte
```diff
@@ -26,5 +26,10 @@
     box-sizing: border-box;
 
     z-index: var(--layer-floating);
+
+    :global(.trakt-action-button) {
+      /** This is required for improved readability when rendering over a cover image */
+      backdrop-filter: blur(var(--ni-16));
```
vladjerca: Let's remove everywhere, we should actually fix the design (contrast?) if it's truly a problem.

PR #1370, opened 2025-12-07
Refactor/search trending

seferturan:
## 🎶 Notes 🎶

- Refactors search to use the server for trending search matches.
- To be followed up by adding the exact matches.

## 👀 Example 👀

https://github.com/user-attachments/assets/5f642985-020e-4522-9521-d169d20978d9

### DIFF
--- a/projects/client/src/lib/utils/array/dedupe.spec.ts
+++ b/projects/client/src/lib/utils/array/dedupe.spec.ts
@@ -1,0 +1,152 @@
+import { describe, expect, it } from 'vitest';
+import { dedupe } from './dedupe.ts';
+
+describe('dedupe', () => {
+  it('should remove duplicates from a single array using string keys', () => {
+    const items = [
+      { id: 'a', value: 1 },
+      { id: 'b', value: 2 },
+      { id: 'a', value: 3 },
+    ];
+
+    const result = dedupe((item) => item.id, items);
+
+    expect(result).toHaveLength(2);
+    expect(result).toEqual([
+      { id: 'a', value: 3 },
+      { id: 'b', value: 2 },
+    ]);
+  });
+
+  it('should remove duplicates from a single array using number keys', () => {
+    const items = [
+      { id: 1, name: 'first' },
+      { id: 2, name: 'second' },
+      { id: 1, name: 'third' },
+    ];
+
+    const result = dedupe((item) => item.id, items);
+
+    expect(result).toHaveLength(2);
+    expect(result).toEqual([
+      { id: 1, name: 'third' },
+      { id: 2, name: 'second' },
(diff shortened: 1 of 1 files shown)


projects/client/src/lib/requests/queries/search/searchTrendingQuery.ts
```diff
@@ -57,13 +59,15 @@ const searchTrendingRequest = (
...
+        ...queryString,
       },
     });
+};
 
 export const searchTrendingQuery = defineQuery({
   key: 'searchTrending',
   invalidations: [],
-  dependencies: (params) => [params.limit],
+  dependencies: (params) => [params.limit, params.query?.toLowerCase().trim()],
```
vladjerca: @seferturan shouldn't we trim before usage?

projects/client/src/lib/utils/array/dedupe.ts
```diff
@@ -0,0 +1,17 @@
+type KeygenValue = string | number;
+type KeygenFunc<T> = (item: T) => KeygenValue;
+
+export function dedupe<T>(
```
vladjerca, line 4: Maybe add a few test cases

projects/client/src/lib/features/search/useSearch.ts
```diff
@@ -71,7 +90,27 @@ export function useSearch() {
...
+      if (currentMode === 'people') {
+        return searchQuery;
+      }
+
+      const trendingQuery = client.fetchQuery(
+        modeToTrendingQuery(term, currentMode),
+      );
+      return combineLatest([searchQuery, trendingQuery]).pipe(
+        map(([searchResults, trendingResults]) => ({
+          type: 'media' as const,
```
vladjerca, line 108: Why do we hardcode to `media`?

vladjerca (changes requested):

projects/client/src/lib/requests/queries/search/searchTrendingQuery.ts
seferturan: Yep 👌  Added another refactor commit to do this for all the queries

projects/client/src/lib/utils/array/dedupe.ts
seferturan: Done 👌

projects/client/src/lib/features/search/useSearch.ts
seferturan: For type inference. `people` are early exited, at this point it's always `media`. When tackling the exact matches I'll see if I can get the typings inferred properly.

vladjerca (approved):

PR #1376, opened 2025-12-08
Fix/infinite query key

seferturan:
## 🎶 Notes 🎶

- This should be the 'real' fix for the `.length` errors like [this one](https://trakt-tv.sentry.io/issues/81804299/?alert_rule_id=260231&alert_type=issue&notification_uuid=296f4465-87c8-43d6-b9b7-841a4d5d7a50&project=4509870926463056).
  - Previously stored query data is incompatible with the new infinite queries. Data in the old format was loaded, tanstack tried to parse `pages` which does not exist in those cases.
    - They were stored in the format of `{ entries: T[]; page: { current: number, total: number}}`
    - For infinite queries, the format should be `{ pages: T[]; pageParams: K[]}`
  - Repro:
    - Clear idb
    - Check out a commit before infinite query was added, visit a page (e.g. `http://localhost:5173/users/me/lists`).
    - Check out `main`, and visit that page again.
    - 💥
- Small fix for `dev mode`; visiting lists page with a non vip account no longer errors.

### DIFF
--- a/projects/client/src/routes/users/[user]/lists/+page.svelte
+++ b/projects/client/src/routes/users/[user]/lists/+page.svelte
@@ -53,16 +53,18 @@
   <PersonalLists slug={$user.slug} type="collaboration" mode={$mode} />
 
   <RenderFor audience="director">
-    <div class="trakt-lists-preview">
-      <ListsHeader title="Smart lists" />
+    <RenderFor audience="vip">
+      <div class="trakt-lists-preview">
+        <ListsHeader title="Smart lists" />
 
-      {#if $mode === "media" || !mode}
-        <SmartListRenderer type="movie" />
-        <SmartListRenderer type="show" />
-      {:else}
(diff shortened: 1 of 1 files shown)


projects/client/src/routes/users/[user]/lists/+page.svelte
```diff
@@ -53,16 +53,18 @@
   <PersonalLists slug={$user.slug} type="collaboration" mode={$mode} />
 
   <RenderFor audience="director">
-    <div class="trakt-lists-preview">
-      <ListsHeader title="Smart lists" />
+    <RenderFor audience="vip">
+      <div class="trakt-lists-preview">
+        <ListsHeader title="Smart lists" />
```
Comment (vladjerca) [inline projects/client/src/routes/users/[user]/lists/+page.svelte:58]: Just realized we haven't translated this 😂

Comment (seferturan) [reply projects/client/src/routes/users/[user]/lists/+page.svelte:58]: Will do it when we make this widely available 👌

PR #1430, opened 2025-12-19
Feat/trivia list

seferturan:
## 🎶 Notes 🎶

- Adds `Trivia` section to movie and show summary.
- For now, long trivias are scrollable in the card itself; will need to be followed up to only show a preview and use a drawer (similar to comments).
- No placeholders atm; the list is not shown if there is no trivia.
- If there are no spoilers in the trivia, the toggles aren't shown.
- Contains a "test feature" to keep the toggle position stable; depending on how this feels in practice I might use this in other places also or change it back.

## 👀 Example 👀

Screenshot 2025-12-19 at 15 05 38


vladjerca (approved):

CanePlayz:
@seferturan Perhaps my question may have been overlooked?

seferturan:
@CanePlayz, ah yeah, sorry 😅 I can't disclose the exact source atm, but it's from a reputable source.

PR #1442, opened 2025-12-22
Perf/paginated comments

seferturan:
## 🎶 Notes 🎶

- Pt1 of comments improvements.
- Comments are now paginated
  - They also have a min-height on the content, resulting in less comment cards being rendered off screen.
- Scrolling an active comment into view is now more robust and properly deals with dynamic heights in the list.
  - Previous version worked 'by accident' because we were fully rendering off screen comments.
- Small fix to invalidations: invalidate comments on reply, since the comments `.replyCount` is used.

## 👀 Example 👀
Before:

https://github.com/user-attachments/assets/92918d18-a358-4730-8274-0be5413b33f3

After:

https://github.com/user-attachments/assets/30b837bd-f048-4a6d-aa8d-4d5345d9f4de

### DIFF
--- a/projects/client/src/lib/requests/models/InvalidateAction.ts
+++ b/projects/client/src/lib/requests/models/InvalidateAction.ts
@@ -28,7 +28,7 @@ type TypeDataMap = {
   'dropped': MediaType;
   'restored': 'show';
   'react': 'comment';
-  'comment': 'reply';
+  'comment_reply': ExtendedMediaType;
   'listed': MediaType;
   'user': UserType;
   'check_in': null;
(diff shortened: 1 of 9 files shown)


projects/client/src/lib/components/card/Card.svelte
```diff
@@ -3,20 +3,25 @@
...
 
   const {
     children,
     variant = "opaque",
     classList = "",
+    action,
   }: ChildrenProps & {
     variant?: "transparent" | "opaque";
     classList?: string;
+    action?: (element: HTMLElement) => void;
```
vladjerca, line 18: Let's follow the same pattern as SMUI for external actions:

https://github.com/hperrin/svelte-material-ui/blob/master/packages%2Fcard%2Fsrc%2FCard.svelte

@seferturan you can refactor this separately, I wanna unblock this since it's perf.

PR #1762, opened 2026-02-27
feat: strings bank sync

michaldrabik:
- sync new batch of strings
- updated escapeXml with some new needed rules. Could be naive so please check.

Will run translations after pre-approvals as usual.

### DIFF
--- a/projects/client/i18n/meta/en.json
+++ b/projects/client/i18n/meta/en.json
@@ -2517,9 +2517,7 @@
     "list_title_liked_lists": {
       "default": "Liked",
       "description": "Title for lists containing the user's liked lists. This title should be as short and concise as possible.",
-      "exclude": [
-        "android"
-      ]
+      "exclude": []
     },
     "text_share_movie": {
       "default": "Just discovered \"{title}\" - you need to see this! 🎬",
@@ -3427,7 +3425,6 @@
       "default": "Now watching",
       "description": "Text for the button that allows users to indicate that they are starting to watch a movie or episode.",
       "exclude": [
-        "android",
         "ios"
       ]
     },
@@ -4056,7 +4053,6 @@
       "default": "There are no shows in this list yet.",
       "description": "Placeholder text shown when there are no shows in a user's personal list.",
       "exclude": [
-        "android",
         "ios"
       ]
     },
@@ -4064,7 +4060,6 @@
       "default": "There are no movies in this list yet.",
       "description": "Placeholder text shown when there are no movies in a user's personal list.",
       "exclude": [
-        "android",
         "ios"
       ]
     },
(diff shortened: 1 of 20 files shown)


projects/client/i18n/meta/en.json
```diff
@@ -7183,6 +7152,302 @@
...
+    "text_onboarding_1_description": {
+      "default": "Track what you're watching and what you want to watch next.\\n\\nPick up right where you left off, save titles for later, and organise everything into smart watchlists or your own personal lists, in one place.",
+      "description": "Text for the description of the first onboarding screen that introduces users to the main value proposition of Trakt.",
+      "exclude": [
+        "web",
+        "ios"
+      ]
+    },
+    "text_onboarding_2_title": {
+      "default": "Find where to watch it.\\nJust press play.",
```
seferturan: Shouldn't these `\\n` ones be 2 entries? This feels like doing formatting in the translations.

projects/client/i18n/meta/en.json
```diff
@@ -7183,6 +7152,302 @@
...
+    "text_onboarding_4_description": {
+      "default": "You've only scratched the surface.\\nJoin Trakt for free and discover more features waiting for you.",
+      "description": "Text for the description of the fourth onboarding screen that introduces users to the main value proposition of Trakt.",
+      "exclude": [
+        "web",
+        "ios"
+      ]
+    },
+    "warning_prompt_mark_as_watched_show_android": {
+      "default": "<![CDATA[Are you sure you want to mark <b><u>all {count} episodes</u></b> of \"{show}\" as watched?]]>",
```
seferturan: This also seems like something that shouldn't be in a translation file 😅

michaldrabik: Not pretty yeah but that's the very Android thing - I even suffixed with **_android**
Basically the most common way (with HTML tags) to apply styling to words in localize way.

You'd just be using the `warning_prompt_mark_as_watched_show` one and iOS I don't recall right now what is the way to do it there : )

projects/client/i18n/meta/en.json
michaldrabik: Hmm yeah. Also very common in Google world if you want to control break lines but in our case it can be split for robustness

projects/client/i18n/meta/en.json
seferturan: What abut the CDATA bit? Can you append that on your end? 🤔

michaldrabik: Apparently no, this needs to sit inside my XML resource afaik :(

michaldrabik: I can try different approach with some custom markers like ** ** inside string and then styling based on that.

PR #1919, opened 2026-03-20
Home watching streak

rudf0rd:

Adds a watching streak callout to the home page that shows authenticated users their current consecutive-day watching streak. The callout displays a tier-based emoji and motivational message that escalates with streak length (Spark at 1 day up to Eternal at 365+), along with an accumulator visualization that breaks the streak down into months, weeks, and days using colored squares with hover tooltips.

When a user hasn't watched anything today but has activity yesterday, the callout shifts to an "at risk" state with a "Watch today to keep your streak alive!" prompt. The accumulator is hidden on mobile, and the streak count stacks to its own line at phone widths for a cleaner 3-line layout.

Streak calculation is a pure function (computeStreak) that builds a Set of day-keys from the user's activity history and walks backwards from today counting consecutive days. All strings are fully localized across 20 locales.

Desktop:
CleanShot 2026-03-20 at 16 26 17@2x

Mobile:
CleanShot 2026-03-20 at 16 26 49@2x

The core computeStreak function is tested in useStreak.spec.ts covering:
- Active streak with consecutive days
- At-risk state when today has no activity but yesterday does
- No streak when there's a gap beyond yesterday
- Single day streaks
- Multiple watches on the same day (deduplication)
- Empty activity history

The function is pure and accepts an injectable now parameter, making all date logic deterministic in tests without mocking.

### DIFF
--- a/projects/client/src/lib/sections/stats/_internal/StreakAccumulator.svelte
+++ b/projects/client/src/lib/sections/stats/_internal/StreakAccumulator.svelte
@@ -31,9 +31,13 @@
   );
 </script>
 
-{#snippet streakGroup(count: number, label: string, type: "month" | "week" | "day")}
+{#snippet streakGroup(
+  count: number,
+  label: string,
+  type: "month" | "week" | "day",
+)}
   {#if count > 0}
-    <Tooltip content={label} variant="compact">
+    <Tooltip content={label}>
       <div class="trakt-streak-group">
         {#each Array.from({ length: count }) as _, i (i)}
           <div
@@ -52,7 +56,7 @@
   {@render streakGroup(dayCount, dayLabel, "day")}
 
   {#if !active}
-    <Tooltip content={m.text_stats_watch_today()} variant="compact">
+    <Tooltip content={m.text_stats_watch_today()}>
       <div class="trakt-streak-square day empty"></div>
     </Tooltip>
   {/if}


projects/client/i18n/meta/fr-fr.json
```diff
@@ -3075,6 +3075,99 @@
...
+    },
+    "text_stats_streak_exact_week": {
+      "default": "A whole week!"
+    },
+    "text_stats_streak_exact_three_days": {
+      "default": "Three days in!"
+    },
+    "text_stats_streak_exact_it_begins": {
+      "default": "It begins!"
+    },
```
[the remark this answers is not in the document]
rudf0rd: why is french always a problem..

projects/client/i18n/meta/en-au.json
```diff
@@ -1654,7 +1654,7 @@
       "default": "Feedback"
     },
     "link_text_terms": {
-      "default": "Terms"
+      "default": "Terms of Use"
```
vladjerca: @rudf0rd Please use the `i18n:traktify` script to add localizations, that doesn't mutate existing values. 

This change for example breaks things for @tysonkerridge on iOS. Please only ever modify `new` values, exiting values should only be altered if the source in `en.json` is touched.

rudf0rd: cleaned.  manually translated french and  hurried a code review :/

projects/client/src/lib/sections/stats/_internal/activityHistoryParams.ts
```diff
@@ -0,0 +1,43 @@
...
+  movieActivityHistoryQuery,
+} from '$lib/requests/queries/users/movieActivityHistoryQuery.ts';
+import {
+  showActivityHistoryQuery,
+} from '$lib/requests/queries/users/showActivityHistoryQuery.ts';
+import { usePaginatedListQuery } from '../../lists/stores/usePaginatedListQuery.ts';
+
+const historyLimit = 1000;
+
+export function useActivityHistory(slug: string) {
```
vladjerca: 2 things here:

1. Filename should technically be `useActivityHistory`
2. Delete this file as it is unused.

projects/client/src/lib/sections/stats/_internal/StreakAccumulator.svelte
```diff
@@ -0,0 +1,150 @@
...
+    <div class="trakt-streak-group">
+      <div class="trakt-streak-tooltip">
+        {monthCount === 1
+          ? m.text_stats_streak_month({ count: String(monthCount) })
+          : m.text_stats_streak_months({ count: String(monthCount) })}
+      </div>
+      {#each Array.from({ length: monthCount }) as _, i (i)}
+        <div class="trakt-streak-square month"></div>
+      {/each}
+    </div>
```
vladjerca: I would appreciate it if we could extract separate components for these:

These are duplicated 3 times but we can extract a local snippet @gemini-code-assist might be able to provide an example of what I mean.

projects/client/src/lib/sections/stats/_internal/StreakAccumulator.svelte
```diff
@@ -0,0 +1,150 @@
...
+  {/if}
+
+  {#if dayCount > 0}
+    <div class="trakt-streak-group">
+      <div class="trakt-streak-tooltip">
+        {dayCount === 1
+          ? m.text_stats_streak_day({ count: String(dayCount) })
+          : m.text_stats_streak_days({ count: String(dayCount) })}
+      </div>
```
vladjerca: Not sure but I see no tootlips? 

Image

projects/client/src/lib/sections/stats/_internal/StreakAccumulator.svelte
```diff
@@ -0,0 +1,150 @@
...
+  .trakt-streak-square {
+    width: var(--ni-20);
+    height: var(--ni-20);
+    border-radius: var(--ni-4);
+    cursor: default;
+    transition: transform var(--transition-increment) ease-in-out;
+    position: relative;
+
+    &.day {
+      background: var(--purple-500);
```
vladjerca: I would enjoy it if we have something that's transparent inside with a full border. A full colored square feels a bit too much 😛

projects/client/src/lib/sections/stats/StreakCallout.svelte
```diff
@@ -0,0 +1,159 @@
...
+    </div>
+
+    <StreakAccumulator days={$streakCount} active={!isAtRisk} />
+  </div>
+{/if}
+
+<style lang="scss">
+  @use "$style/scss/mixins/index" as *;
+
+  .trakt-streak-callout {
```
vladjerca, line 164: I think here we can also leverage the `Card` Component for a more normalized behavior @seferturan we might have other components too that we can re-use (for example the ones we use to announce, like the Banners?)

Light theme is again broken: 

Image

projects/client/src/lib/sections/stats/_internal/StreakAccumulator.svelte
rudf0rd: good catch.  tooltip broken when no streak.

rudf0rd:
## Fresh

### Desktop

#### Dark
image

#### Light
CleanShot 2026-03-24 at 19 22 37@2x


---

### Mobile

#### Dark
CleanShot 2026-03-24 at 19 20 44@2x

#### Light
CleanShot 2026-03-24 at 19 23 00@2x

projects/client/src/lib/sections/stats/_internal/StreakAccumulator.svelte
rudf0rd: I actually use the empty square to represent a `Continue your streak` moment.  I did tone down the purple.

Side note: I think instead of a skeleton on this one, we add an empty state, but wanted to confirm that.  Making something persistent could be annoying if unwanted.

projects/client/src/lib/sections/stats/_internal/useActivityHistory.ts
```diff
@@ -0,0 +1,60 @@
...
+    startDate,
+    endDate,
+  };
+
+  const {
+    list: movies,
+    isLoading: isLoadingMovies,
+    hasNextPage: hasNextMoviePage,
+    fetchNextPage: fetchNextMoviePage,
+  } = usePaginatedListQuery(movieActivityHistoryQuery(params));
```
vladjerca: @rudf0rd the fact that we just paginate the entire user history here for the past year is really bad in terms of perf.

We need to use the minimal history which we already load in order to display states in the client (watched, etc.)

vladjerca (changes requested):

projects/client/src/lib/sections/stats/_internal/StreakAccumulator.svelte
vladjerca: @rudf0rd I updated this here https://github.com/trakt/trakt-web/pull/1919/commits

projects/client/src/lib/sections/stats/_internal/useActivityHistory.ts
seferturan: This now fully goes through the minimal user history, this file has been deleted 👌

PR #1948, opened 2026-03-23
iOS i18n updates + overall sentiment analysis translations

tysonkerridge:
- Added overall sentiment analysis result keys.
- Enabled a bunch of other keys for iOS usage.


tysonkerridge:
@gemini-code-assist your latest review comments regarding missing `exclude` properties are unnecessary. The `exclude` property only needs to be in the `en.json` file. You can see how all the other files don't have the `exclude` property.

PR #1974, opened 2026-03-27
Show: total runtime, aired episode count

rectifyer:
## Description

This feature introduces the display of a show's total runtime and episode count within the media details section. It aims to provide users with a more comprehensive overview of a TV show's overall length, enhancing the information available at a glance.

The changes include:
- Expanding the `ShowEntry` data model to support a `totalRuntime` field.
- Implementing logic to map `total_runtime` from the API or calculate it based on individual episode runtime and aired episode count.
- Adding new internationalization (i18n) keys for "Total Runtime" and for singular and plural forms of "episode count" across all supported languages.
- Developing UI components to format and display the total runtime, along with the formatted episode count, for TV shows.

## Screenshots

image

## Testing

New unit tests have been added to `MediaDetails.spec.ts` to verify the correct display of the total runtime for a show and to ensure it is hidden when data is unavailable (e.g., `totalRuntime` is `NaN`).

### DIFF
--- a/projects/client/src/lib/requests/models/Season.ts
+++ b/projects/client/src/lib/requests/models/Season.ts
@@ -14,5 +14,6 @@ export const SeasonSchema = z.object({
   airDate: z.date(),
   rating: z.number().nullish(),
   network: z.string().nullish(),
+  totalRuntime: z.number(),
 });
 export type Season = z.infer<typeof SeasonSchema>;
--- a/projects/client/src/lib/requests/queries/shows/showSeasonsQuery.ts
+++ b/projects/client/src/lib/requests/queries/shows/showSeasonsQuery.ts
@@ -43,6 +43,7 @@ export const mapToSeason = (item: SeasonsResponse[0]): Season => {
     airDate: new Date(item.first_aired ?? MAX_DATE),
     rating: mapToTraktRating(item.rating),
     network: item.network,
+    totalRuntime: item.total_runtime ?? NaN,
   };
 };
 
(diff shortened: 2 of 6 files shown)


projects/client/src/lib/sections/summary/components/details/_internal/useMediaDetails.ts
```diff
@@ -78,6 +79,48 @@ function networks(entries: MediaNetwork[] | undefined) {
...
+}
+
+function totalRuntime(show: ShowEntry): MediaDetail {
+  const totalMinutes = show.totalRuntime;
+
+  if (!Number.isFinite(totalMinutes) || totalMinutes <= 0) {
+    return { title: m.header_total_runtime() };
+  }
+
+  const duration = formatTotalDuration(totalMinutes);
```
seferturan: I think here we can directly use `const duration = toHumanDuration({ minutes: totalMinutes }, languageTag());`?

projects/client/src/lib/sections/summary/components/details/_internal/useMediaDetails.ts
```diff
@@ -78,6 +79,48 @@ function networks(entries: MediaNetwork[] | undefined) {
...
+  const count = show.episode.count;
+  const episodes = count === 1
+    ? m.text_episode_count_one({ count })
+    : m.text_episode_count_other({ count });
```
seferturan: Maybe episode count can be it's own thing with header `Episode Count`, then we also won't have to deal with translations of plural vs singular like this.

projects/client/src/lib/sections/summary/components/details/_internal/useMediaDetails.ts
rectifyer: Good idea, also fixes the above Gemini comment.

projects/client/src/lib/sections/summary/components/details/_internal/useMediaDetails.ts
rectifyer: I didn't want to add 2 new entries is mainly why I combined it in one.

seferturan (approved):

PR #1982, opened 2026-03-30
WIP: progress page

### LINKED ISSUE #1402: Bring back the progress page
v3 currently has a lot fewer pages than v2. While I believe that some will come back, I do assume you're also using v3 to restructure pages internally and perhaps get rid of some. One page I'm missing dearly is the progress page. Progress bars, percentages, seeing all watched shows, it's really great.

Or as user "TopCat" on the forums has put it:

> The Progress page is the one place on the website you can find all of the shows that you’ve ever watched in order, easily see season by season progress, maybe you’re looking into rewatching something of old that you’ve already seen - it’s easily accessible with only a few clicks.
> 
> The sorting and filtering on the Progress page is also very useful, yes, as is the progress bars and the “Watched percentage”, Time Left To Watch and Total Amount of Shows at the top of the page.
> 
> It’s all in the details, at least for me. Such a page is a blessing in disguise.

And here's a comment from another user:

> The complete absence of the Progress Page, which is my most visited, is absurd because I like seeing more direct details than Up Next provides. I like being able to instantly sort so I can see what’s next, but also then get a quick idea of what has the least time, change the sorts and see views without having to go deep into Settings to change my Up Next for two seconds and then change it back to what I’d like for my Default.

rectifyer:
## 🚨 WIP

image

### DIFF
--- a/progress-variations.html
+++ b/progress-variations.html
@@ -1,0 +1,2517 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Trakt — Watched Progress Design Variations</title>
+
+  <!-- Google Fonts -->
+  <link rel="preconnect" href="https://fonts.googleapis.com">
+  <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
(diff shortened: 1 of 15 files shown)


nFin1ty:
> We have a different design for progress… IMG_1487

...which looks so bad and provides little to none usability. Can't compare at all with the old Progress page.

PR #1984, opened 2026-03-30
Chore/discover param

### LINKED ISSUE #1473: Discover toggles should be reflected in the URL
These toggles:

Image

The state is now in a store, switch to using search params.

seferturan:
## 🎶 Notes 🎶

- Fixes #1473
- Discover mode is now reflected in search params
- Also contains a small fix where search mode toggles would override global search params.

## 👀 Example 👀

https://github.com/user-attachments/assets/d0708163-6044-49d0-a9ea-b64ede0d7a2e

### DIFF
--- a/projects/client/src/lib/sections/lists/user/UserListPaginatedList.svelte
+++ b/projects/client/src/lib/sections/lists/user/UserListPaginatedList.svelte
@@ -6,20 +6,18 @@
   import DrilledMediaList from "../drilldown/DrilledMediaList.svelte";
   import SortValue from "./_internal/SortValue.svelte";
   import UserListItem from "./_internal/UserListItem.svelte";
-  import ListActions from "./ListActions.svelte";
   import type { SortBy } from "./models/SortBy";
   import type { SortDirection } from "./models/SortDirection";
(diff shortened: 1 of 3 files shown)


projects/client/src/lib/features/discover/DiscoverProvider.svelte
```diff
@@ -1,18 +1,48 @@
...
   import { createDiscoverContext } from "./_internal/createDiscoverContext";
+  import type { DiscoverMode } from "./models/DiscoverMode";
 
   const { children }: ChildrenProps = $props();
 
-  const { current } = useToggler("discover");
+  const { current, options } = useToggler("discover");
+  const { update: updateParameters } = useParameters();
+
+  const rawUrlMode = page.url.searchParams.get(DISCOVER_MODE_PARAM);
```
vladjerca: I don't like this, the source of truth should be the URL.

Clicking on the media switcher should be a hyperlink and that should trigger everything else.

seferturan: Done. That simplified things greatly 😅

PR #2021, opened 2026-04-02
refactor: exclude currently unused Android strings

michaldrabik:
Just some clean ups.

### DIFF
--- a/projects/client/i18n/meta/en.json
+++ b/projects/client/i18n/meta/en.json
@@ -73,6 +73,7 @@
       "default": "Recommended Movies",
       "description": "Title for lists containing recommended movies. This title should be as short and concise as possible.",
       "exclude": [
+        "android",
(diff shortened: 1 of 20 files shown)


projects/client/i18n/meta/en.json
```diff
@@ -2170,20 +2181,23 @@
...
       "exclude": [
+        "android",
         "ios"
       ]
     },
     "text_reply_by": {
       "default": "Reply by",
       "description": "Text used to indicate the author of a reply to a review.",
       "exclude": [
+        "android",
```
seferturan: I think all the entries that now exclude all clients can be completely removed. And are gemini's comments relevant to your situation?

michaldrabik: Double checked gemini and yeah - removed those fully excluded atm.

PR #2054, opened 2026-04-08
fix(i18n): add missing exclude options for recent strings

michaldrabik:
- adds missing excludes for recently added strings


tysonkerridge:
@michaldrabik I just removed iOS from a few that are already being used but I hadn't synced here yet

michaldrabik:
> @michaldrabik I just removed iOS from a few that are already being used but I hadn't synced here yet

Yeah thanks thats why I have @ you :)

vladjerca (approved):

PR #2077, opened 2026-04-10
feat(ci): connect and enable crowdin in our main ci_cd

michaldrabik:
- this PR prepares main flow to update translations for any changes that might have occurred in the source **en.json**
- bonus: current translations were just sorted by key name

On push to main only,
if anything changes in our source meta file, the CrowdIn is updated with latest source files.

The AI pre-translation is applied and result translations bundles are downloaded and pushed without re-triggering the flow.


projects/client/i18n/meta/en.json
```diff
@@ -33,7 +33,7 @@
       ]
     },
     "header_profile_banner_greeting": {
-      "default": "Hello, {name}",
+      "default": "How are you, {name}",
```
seferturan: These were test changes you made right? 😅

seferturan (changes requested):

.github/workflows/ci_cd.yml
seferturan, line 127: By default, this takes the ref/SHA that triggered the workflow, so the changes pushed to main by `localize` won't be taken here.

For all cases after localize maybe:
````
- uses: actions/checkout@v4
  with:
    ref: main
````
Though that's also not very safe. Can we get the sha of the commit from the localize action? 🤔

seferturan (changes requested):

projects/client/i18n/meta/en.json
michaldrabik: I am actually gonna remove that one altogether - its not used anymore.

.github/workflows/ci_cd.yml
michaldrabik: I see you. Unfortunately the action does not seem to be outputting SHA to be used (https://github.com/marketplace/actions/crowdin-action)

So this leaves 2 options:
- one you proposed with ref: main (i see the concern here but its not that high risk is it?)
- use alternative approach idea -> whenever push to main happens, could trigger 2 flows `ci_cd` and extract `localize` to separate one. As soon as `ci_cd` detects that translations are needed it will finish. The `localize` will perform needed steps and re-trigger `ci_cd` after translations are pushed. This would re-trigger `ci_cd` with all changes since the 1st step.

michaldrabik: Actually even easier for `#2` idea - do not trigger at the same time in parallel. Just let ci_cd detect new translations and it can trigger localize workflow and finish. localize will perform needed steps and re-trigger ci_cd after translations are pushed.

michaldrabik: Alternative workflow now proposed with latest commit.

seferturan (approved):

PR #2115, opened 2026-04-15
Year in Review: default template

### LINKED ISSUE #2007: Migrate YIR templates

### LINKED ISSUE #2010: Migrate pre 2020 template
- The initial version can focus on the non chart content.
- In this version the charts can have placeholder components (which can then be implemented later on)
  - e.g. `lib/components/charts/bar-chart/BarChart.svelte`,  `lib/components/charts/circle-pack/CirclePack.svelte`, etc.

### LINKED ISSUE #2011: Add bar chart
To be used in the yir/mir templates.
The earlier created wrapper component can be implemented using Carbon Charts (https://charts.carbondesignsystem.com/bar)

E.g. to recreate these from the 2019 template:

Image

### LINKED ISSUE #2012: Add circle pack chart
To be used in the yir/mir templates.
The earlier created wrapper component can be implemented using Carbon Charts (https://charts.carbondesignsystem.com/circlepack); for studios/networks

E.g. to recreate these from the 2019 template:

Image

### LINKED ISSUE #2013: Add "meter" chart
Not sure about the naming here, but the one used for genres :P 
To be used in the yir/mir templates.

Might be possible to use https://charts.carbondesignsystem.com/meter for this

E.g. to recreate these from the 2019 template:

Image

rectifyer:
## Description

Introduces the **Year in Review (YIR) default template** — a full-page, visually rich summary of a user's watching activity for a given year. The feature is gated behind the `YearInReview` feature flag and renders at `/users/[user]/year/[year]`.

## Screenshots

image

image

image

## What's included

**Page composition & layout**
- `YirPage` entry point with loading state, delegating to `YirDefault` — a composition shell that conditionally renders each section based on available data
- `TraktPage` updated with a `content-only` mode to allow full-bleed rendering without standard content wrapper spacing

**Sections** (each under `lib/sections/yir/default/_internal/`):
- **Title** — Full-viewport hero with user avatar, display name, year, and cover image background
- **Totals** — Aggregate stats: total plays, hours watched, collected, ratings, comments, lists
- **Calendar** — First/last watched item of the year with fanart background and formatted date
- **Stats** — Detailed hours/plays breakdown (total, per-month, per-week, per-day) with four chart visualizations (weekly, monthly, daily, hourly)
- **Most Watched** — Ranked carousel with fanart crossfade on hover and thumbnail grid
- **Genres** — Horizontal stacked bars (desktop) / vertical list (mobile) with color-coded genre breakdown
- **Networks / Studios** — Circle-pack bubble charts via Carbon Charts with injected company logos and brand colors
- **Rated** — Grid of top-rated posters with rating badges and fanart backgrounds
- **People** — Paginated grid of actors, actresses, directors, writers with circular headshots and hover tooltips
- **Upgrade CTA** — VIP upsell shown only to free users via `RenderFor`

**Charts** (Carbon Charts):
- `YirWeeklyPlaysChart` (52 bars), `YirMonthlyPlaysChart` (12 bars), `YirDailyPlaysChart` (7 bars), `YirHourlyPlaysChart` (24 bars) — all with locale-aware labels and peak-bar highlighting
- `YirNetworksChart` — Circle-pack with `MutationObserver`-driven logo injection

**API layer:**
- `yirDetailQuery` + `mapToYirDetail` — fetches and maps the full YIR response with Zod-validated domain models
- `yirPeopleQuery` + `mapToYirPerson` — fetches people data by type (actors, actresses, directors, writers)
- Reactive hooks: `useYirDetail`, `useYirPeople`

**Utilities:**
- `findMaxIndex` — returns index of max value in an array (used by charts)
- `formatNumber` — locale-aware number formatting with rounding
- `appendWebp` — appends `.webp` to image URLs for optimized delivery

**Other changes:**
- Added `YearInReview` to `FeatureFlag` enum
- `prependHttps` now normalizes legacy CDN hostnames (`walter-r2.trakt.tv`, `walter.trakt.tv`) to `media.trakt.tv`
- `userProfileQuery` extended to request image data for cover images
- `mapToUserProfile` applies WebP optimization to VIP cover images
- 289 new i18n keys for all YIR labels, section titles, and UI states

## Tests

- `findMaxIndex.spec.ts` — 9 test cases covering edge cases (ties, negatives, single element, decimals)
- `formatNumber.spec.ts` — 8 locale-agnostic test cases covering rounding, separators, and negative numbers

### DIFF
--- a/projects/client/i18n/meta/en.json
+++ b/projects/client/i18n/meta/en.json
@@ -7220,6 +7220,432 @@
         "ios"
       ]
     },
+    "yir_title_year_to_date": {
+      "default": "Year to date",
+      "description": "Title shown for the current year's Year in Review.",
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_title_year_in_review": {
+      "default": "Year in Review",
+      "description": "Main title for the Year in Review page.",
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_section_title_totals": {
+      "default": "{year} Totals",
+      "description": "Section title for the yearly totals overview.",
+      "variables": {
+        "year": {
+          "type": "number"
+        }
+      },
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_section_title_tv_shows": {
+      "default": "TV Shows",
+      "description": "Section title for TV shows statistics.",
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_section_title_movies": {
+      "default": "Movies",
+      "description": "Section title for movies statistics.",
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_section_title_highest_rated_shows": {
+      "default": "Highest Rated TV Shows",
+      "description": "Section title for highest rated TV shows.",
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_section_title_highest_rated_movies": {
+      "default": "Highest Rated Movies",
+      "description": "Section title for highest rated movies.",
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_section_title_show_genres": {
+      "default": "TV Show Genres",
+      "description": "Section title for TV show genres.",
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_section_title_movie_genres": {
+      "default": "Movie Genres",
+      "description": "Section title for movie genres.",
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_section_title_networks": {
+      "default": "TV Show Networks",
+      "description": "Section title for TV show networks.",
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_section_title_studios": {
+      "default": "Movie Studios",
+      "description": "Section title for movie studios.",
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_section_title_top_actors": {
+      "default": "Top Actors",
+      "description": "Section title for top actors.",
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_section_title_top_actresses": {
+      "default": "Top Actresses",
+      "description": "Section title for top actresses.",
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_section_title_top_directors": {
+      "default": "Top Directors",
+      "description": "Section title for top directors.",
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_section_title_top_writers": {
+      "default": "Top Writers",
+      "description": "Section title for top writers.",
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_calendar_first_play": {
+      "default": "first play of {year}",
+      "description": "Label for the first play of the year in the calendar section.",
+      "variables": {
+        "year": {
+          "type": "number"
+        }
+      },
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_calendar_last_play": {
+      "default": "last play of {year}",
+      "description": "Label for the last play of the year in the calendar section.",
+      "variables": {
+        "year": {
+          "type": "number"
+        }
+      },
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_label_most_watched": {
+      "default": "most watched {type}",
+      "description": "Label for most watched items.",
+      "variables": {
+        "type": {
+          "type": "string"
+        }
+      },
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_label_hours_watched": {
+      "default": "hours watched",
+      "description": "Label for hours watched statistic.",
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_label_per_month": {
+      "default": "per month",
+      "description": "Label for per month statistics.",
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_label_per_week": {
+      "default": "per week",
+      "description": "Label for per week statistics.",
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_label_per_day": {
+      "default": "per day",
+      "description": "Label for per day statistics.",
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_label_collected": {
+      "default": "collected",
+      "description": "Label for collected items statistic.",
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_label_plays_by_week": {
+      "default": "{type} plays by week",
+      "description": "Label for plays by week chart.",
+      "variables": {
+        "type": {
+          "type": "string"
+        }
+      },
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_label_plays_by_month": {
+      "default": "{type} plays by month",
+      "description": "Label for plays by month chart.",
+      "variables": {
+        "type": {
+          "type": "string"
+        }
+      },
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_label_plays_by_day": {
+      "default": "{type} plays by day of the week",
+      "description": "Label for plays by day of week chart.",
+      "variables": {
+        "type": {
+          "type": "string"
+        }
+      },
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_label_plays_by_hour": {
+      "default": "{type} plays by time of day",
+      "description": "Label for plays by time of day chart.",
+      "variables": {
+        "type": {
+          "type": "string"
+        }
+      },
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_label_week_number": {
+      "default": "Week {weekNumber}",
+      "description": "Label for week number in charts.",
+      "variables": {
+        "weekNumber": {
+          "type": "number"
+        }
+      },
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_unit_play": {
+      "default": "{count, plural, one {play} other {plays}}",
+      "description": "Unit for plays (singular/plural).",
+      "variables": {
+        "count": {
+          "type": "number"
+        }
+      },
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_unit_hour": {
+      "default": "{count, plural, one {hour} other {hours}}",
+      "description": "Unit for hours (singular/plural).",
+      "variables": {
+        "count": {
+          "type": "number"
+        }
+      },
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_unit_rating": {
+      "default": "{count, plural, one {rating} other {ratings}}",
+      "description": "Unit for ratings (singular/plural).",
+      "variables": {
+        "count": {
+          "type": "number"
+        }
+      },
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_unit_comment": {
+      "default": "{count, plural, one {comment} other {comments}}",
+      "description": "Unit for comments (singular/plural).",
+      "variables": {
+        "count": {
+          "type": "number"
+        }
+      },
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_unit_list": {
+      "default": "{count, plural, one {list} other {lists}}",
+      "description": "Unit for lists (singular/plural).",
+      "variables": {
+        "count": {
+          "type": "number"
+        }
+      },
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
+    "yir_unit_movie": {
+      "default": "{count, plural, one {movie} other {movies}}",
+      "description": "Unit for movies (singular/plural).",
+      "variables": {
+        "count": {
+          "type": "number"
+        }
+      },
+      "exclude": [
+        "android",
+        "ios"
+      ]
+    },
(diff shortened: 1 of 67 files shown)


projects/client/i18n/meta/en.json
```diff
@@ -6936,6 +6936,291 @@
         "ios"
       ]
     },
+    "yir_title_year_to_date": {
+      "default": "Year to date",
+      "description": "Title shown for the current year's Year in Review.",
+      "exclude": []
```
seferturan: I think for all of these we can exclude `android` and `ios`

projects/client/src/lib/sections/yir/default/_internal/_shared.scss
```diff
@@ -0,0 +1,89 @@
+@use "$style/scss/mixins/index" as *;
+
+// Common section header styles
+.yir-section-header {
```
seferturan: Instead of having this shared scss, these should probably be separate components that have the respective style, e.g. `YirSectionHeader.svelte` etc.

projects/client/src/lib/sections/yir/default/_internal/_shared.scss
```diff
@@ -0,0 +1,89 @@
...
+  @include for-mobile {
+    max-width: 100%;
+  }
+}
+
+// Bubble chart tooltip styles (shared between networks and studios)
+:global(.cds--tooltip),
+:global(.cds--cc--tooltip),
+:global(.cds--cc--tooltip-container) {
+  background: transparent !important;
```
seferturan: Please no `!important`s 😅 It overrides the natural cascade of css.

projects/client/src/lib/sections/yir/default/_internal/YirCalendarSection.svelte
```diff
@@ -0,0 +1,214 @@
...
+    how,
+    item,
+    year,
+  }: {
+    how: "first" | "last";
+    item: YirWatchedItem;
+    year: number;
+  } = $props();
+
+  const watchedDate = $derived(new Date(item.watchedAt));
```
seferturan: Shouldn't `item.watchedAt` always be a date? If so, it can be made a date in the mapper.

projects/client/src/lib/sections/yir/default/_internal/YirCalendarSection.svelte
```diff
@@ -0,0 +1,214 @@
...
+      year: "numeric",
+    }),
+  );
+
+  const formattedTime = $derived(
+    watchedDate.toLocaleTimeString(languageTag(), {
+      hour: "numeric",
+      minute: "2-digit",
+    }),
+  );
```
seferturan, line 27: If none of the current utils in `date/utils` is usable for this, maybe these can be extracted to go there so we can more easily reuse them :D

projects/client/src/lib/sections/yir/default/_internal/_shared.scss
```diff
@@ -0,0 +1,89 @@
...
+// Common section header styles
+.yir-section-header {
+  text-align: center;
+  padding: 70px 0 20px 0;
+
+  h2 {
+    text-transform: uppercase;
+    display: inline-block;
+    letter-spacing: 1px;
+    border: solid 1px #fff;
```
seferturan: This holds for other files also; it makes sense that these YIR templates have their own specific paddings/font-sizes, etc.
But, can we please use existing variables for things like these? E.g. `--ni-*`, `--gap-*`, `--border-radius-*`. Same for most colors, I think most of them can be expressed in therms of `--shade-*`?

seferturan (changes requested):
Well, that's a good match on what it's like on OG :D 

I haven't gone over everything yet, but some remarks for now:

- It seems like this is very close to being light-them compatible :D If most of the background/text colors etc were replaced with the existing background/text-color css variables, does it work?

- This doesn't look very appealing 😅 Maybe we could already show the header section of it already and show a loading indicator there? I'm assuming that everything is preloaded, and we can't do lazy loading as users scroll through it?
Image

- Not all images seem to load.
Image

- Could use some breathing room in for the content. 😅 `--layout-distance-side` can be used for the sides. And on the left an extra `--layout-sidebar-distance` to account for the sidebar. Backgrounds going under the sidebar is fine I'd say 🤔
Image

- This chart has some weird behavior. On larger screens also. It appears 'correct' after either a delay or when I hover.
Image

projects/client/i18n/meta/en.json
rectifyer: Fixed in commit `feat(yir): review 001`

projects/client/src/lib/sections/yir/default/_internal/YirCalendarSection.svelte
rectifyer: Fixed in commit `feat(yir): review 001`

projects/client/src/lib/sections/yir/default/_internal/_shared.scss
rectifyer: Fixed in commit `feat(yir): review 001`

projects/client/src/lib/sections/yir/default/_internal/_shared.scss
rectifyer: Fixed in commit `feat(yir): review 002`

projects/client/src/lib/sections/yir/default/_internal/YirCalendarSection.svelte
rectifyer: Fixed in commit `feat(yir): review 005`

projects/client/src/lib/sections/yir/default/_internal/_shared.scss
rectifyer: Fixed in commit `feat(yir): review 005`

rectifyer:
@seferturan I address all your comments. Also added a fixed top navigation header to match what v2 does. Hopefully the bubble charts load better now, redid some stuff to be more responsive to window size changes too.

I will squash to a single commit once ready to go.

projects/client/src/lib/sections/yir/default/_internal/YirHourlyPlaysChart.svelte
```diff
@@ -0,0 +1,123 @@
...
+  }: {
+    data: number[];
+  } = $props();
+
+  let containerWidth = $state(1170);
+  let chartContainer: HTMLDivElement;
+
+  const barWidth = $derived(Math.ceil((containerWidth - 86) / 24));
+
+  $effect(() => {
```
seferturan: Holds for other charts also:

- We should have our own wrappers for the charts, e.g. our own `BarChart.svelte` which then use the carbon charts one. That way we can keep all internal configs hidden from the other components, and only worry about specific input props we care about. If in the future we ever want to change something, we'll only have to do it in one place. And should also greatly reduce the duplication between `YirDailyPlaysChart`, `YirMonthlyPlaysChart`, `YirWeeklyPlaysChart`, and `YirHourlyPlaysChart`
  - Then the yir charts will only have to deal with mapping the data and such.
- I'm not sure if all these `$effect()` usages are needed, doesn't carbon charts support that out of the box? E.g. their `resizable` prop?
- Also for the colors used in the charts, can we use our color variables?

projects/client/src/lib/sections/yir/default/_internal/YirNetworksChart.svelte
```diff
@@ -0,0 +1,365 @@
...
+    const maxRetries = 10;
+    const tryAddLogos = () => {
+      const success = addLogosAndText();
+      if (!success && retryCount < maxRetries) {
+        retryCount++;
+        setTimeout(tryAddLogos, 200);
+      }
+    };
+
+    setTimeout(tryAddLogos, 100);
```
seferturan: This seems a bit hacky and fragile 😅 If carbon charts doesn't support adding content to the circles, it might be better to render our own using `d3 pack` or something.

seferturan (changes requested):

projects/client/src/lib/sections/yir/default/_internal/YirHourlyPlaysChart.svelte
rectifyer: I will make a shareable BarChart component to re-use and vars for the colors. The reason there is custom effect stuff is related to the spacing in the bar charts. Without it, there are large gaps and I couldn't find a way to get them matching the original YIR design without some customization. I will take another look, but so far this was the way I found to match the design.

projects/client/src/lib/sections/yir/default/_internal/YirNetworksChart.svelte
rectifyer: Will take a look. This was a bit weird to get working consistently which is why I ended up with the code like this. Similar issue with the bubble chart options not fully doing what I wanted. I will take a look at a custom d3 option as an alternative.

rectifyer: Updates in https://github.com/trakt/trakt-web/pull/2115/commits/ff95bede5f05ec0a44343f612132b3b44781e987

projects/client/src/lib/sections/yir/default/_internal/YirHourlyPlaysChart.svelte
rectifyer: Updates in https://github.com/trakt/trakt-web/pull/2115/commits/f9fccbcee28c3a7136f2cdc780a484ee35e9ce57

rectifyer:
@seferturan A summary of what my last commit changed...

- **Shared tooltip** — new `YirTooltip.svelte` + `yirTooltipHTML.ts`. Three-line API (`main` / optional `sub` / optional `extra`); used as a Svelte component for the bubble chart and as an HTML helper for the bar chart.
- **BarChart** — `tooltipFn` → generic `tooltipHTML: (args) => string`; fixed crash by reading `point.label` (matches `mapsTo`) instead of `point.key`; flatten Carbon's wrapper to remove the white outline.
- **YIR page shell** — moved chart/text color overrides and red bar-hover onto `.yir-page` (so all charts inherit, not just Weekly); fixed asymmetric gap around the side navbar.
- **Force-collapsed navbar on YIR** — added `forceCollapsed` to navbar state; YIR sets it on mount, restores on unmount.
- **Most-watched strip** — lifts ~50% over the fanart with rounded outer corners + drop-shadow; dark backdrop so hover-faded thumbs don't bleed the fanart; on small screens it's full-bleed and scrolls horizontally.
- **Bar-chart consumers** — Weekly/Monthly/Daily/Hourly all use the new tooltip helper; Weekly shows 3 lines, others 2.
- **Stats section** — `1fr 1fr` → `minmax(0, 1fr) minmax(0, 1fr)` so charts shrink with the viewport (Carbon's pixel width was pinning the column min-size); added breathing room on small screens.
- **Rated corner rating** — triangle and font scaled down, span position recentered.
- **Header scrolled state** — uses design tokens instead of inline values.
- **Movie thumb mapper bug** — `mapToMovieEntry` now consumes `images.thumb[]` like the show mapper, so Most Watched Movies gets logo-over-fanart thumbs.

rectifyer:
@seferturan Details on my recent fix up commit. I think should address everything we talked about, even moving the genres into a shareable component for the future.

## YIR default template — fixup summary

### Responsive bottom-axis tick labels (`BarChart`)
- Added `minTickWidth` prop (`BarChartProps.ts`, default 20) — minimum px per visible bottom-axis tick.
- New `responsiveTickValues` derived: when `containerWidth / data.length < minTickWidth`, drops labels at a fixed stride `ceil(minTickWidth / slot)`.
- Output is always padded to `data.length` with `""` for skipped slots — works around lodash `_.merge` in Carbon's `model.setOptions`, which merges arrays *by index* and would otherwise leave stale tail labels in place when the viewport widens.
- Pinned `ticks.number = values.length` to suppress Carbon's overlap-driven auto-thinning, which otherwise reinjects extra ticks on top of our subset.
- Hourly chart now passes `minTickWidth={40}` since "12 AM"-style labels are wider than week numbers.

### `RankedBars` chart component (extracted from `YirGenresSection`)
- New `$lib/components/charts/RankedBars.svelte` + `models/RankedBarsProps.ts`.
- Generic API: `items: { name, count }[]`, `total`, `itemLabel`, optional `colors[]` (default palette built in).
- Behavior preserved verbatim from old genres section: alternating up/down tier labels, hover percentage badge, mobile vertical stack at `for-tablet-sm-and-below`, `200%`-width labels with `text-overflow: ellipsis` so a label can use its bar + neighbor.
- `YirGenresSection` is now a thin wrapper: section chrome (`<section>` / `YirPageInner` / `YirSectionHeader`) + `<RankedBars>`. Reusable across other YIR templates and an all-time view.

### Placeholder image handling
- `YirCalendarSection`, `YirMostWatchedSection`, `YirPeopleGroup` now check `PLACEHOLDERS.includes(url)` instead of a falsy URL check, so backend-supplied placeholder URLs are detected.
- `YirPeopleGroup` falls back to `DEFAULT_AVATAR` when the headshot is a placeholder, with an `is-default` class that resets the `top: -10%` offset (default avatars don't need the crop adjustment).

### Stats section layout
- Added `<ArrowRightIcon>` between each stat in both rows; first arrow is `visibility: hidden` so columns align identically across the two rows without an extra spacer.
- Switched `align-items: center` → `flex-start` and computed arrow `padding-top: var(--ni-14)` so arrows sit on the big-number baseline instead of the row centerline. Smaller `--ni-6` top padding at `tablet-sm-and-below`.
- Hardcoded font sizes replaced with tokens: `60/45/30px` → `--ni-48/--ni-40/--ni-30`; unit label `13px` → `--ni-12`.

### People count layout
- Movie/show counts now always render two `<h3>` lines per person (using `&nbsp;` when zero) so cards have a consistent height and the grid doesn't snap when one count is missing.

### Misc
- `YirMostWatchedSection` title font: `40px` / `32px` (mobile) → `--ni-32` / `--ni-24` tokens.

seferturan (approved):
Some small extra changes:

- I inlined the genres chart again; would require some more changes to make it generic. We can do that as a follow up.
- I reverted the changes for mapping the movie thumb. This changed the images for movies everywhere; we can tackle this as a follow up.
- I changed the behavior of the ticks in the charts; consumers should pass in the tick labels separately, having a min width there isn't really scalable. The labels shown should still make sense (e.g. when limit week numbers, show a number every 4 weeks etc).
- I replaced all font sizes with numeric increments.

PR #2270, opened 2026-05-06
iOS i18n updates

tysonkerridge:
added some more translations used in tvOS and enabled some existing ones.

@michaldrabik I updated the "button_text_player_resume" value as it had "Play Now" instead of "Resume"


michaldrabik:
> I updated the "button_text_player_resume" value as it had "Play Now" instead of "Resume"

@tysonkerridge Yeah it was mistake I had it already pending in my other PR but removed it now on my side to avoid duplication.

seferturan (approved):

PR #2329, opened 2026-05-14
feat: update weekly pulse stats graphs and streak UI

ElMagnea:
Improves weekly pulse statistics with enhanced graphs and streak accumulator UI.


seferturan:
Will be tackled differently (e.g. when we add the screen time version).

PR #2330, opened 2026-05-14
feat: improve list filter and dropdown navigation

ElMagnea:
Enhances list filtering and dropdown navigation components.


seferturan:
Closing this. Will be tackled differently. Also this was missing the select component 😅

PR #2331, opened 2026-05-14
feat: update homepage layout and VIP theme colors

ElMagnea:
Updates homepage layout by removing WeeklyPulse component and adjusts VIP theme colors.


seferturan:
Will be tackled differently (e.g. when we add the screen time version).

PR #2353, opened 2026-05-15
fix: guard show summary against transient nullish data

vladjerca:
## Summary
- Tighten the show-page \`isReady\` predicate to require \`$show\`, \`$intl\`, \`$studios\`, \`$crew\`, **and \`$seasons\`** alongside \`!\$isLoading\`
- Drop the \`!\` non-null assertions from the prop pass to \`<ShowSummary>\`
- \`SeasonList\` keeps its non-nullable \`Season[]\` contract

## Why
The route gated rendering on \`!\$isLoading\` only and then asserted every store non-null with \`!\`. During a navigation transition the query can settle with \`isLoading=false\` before all derived stores repopulate — \`undefined\` reached \`SeasonList\` and crashed at \`seasons.length\`.

- [TRAKT-WEB-87F](https://trakt-tv.sentry.io/issues/TRAKT-WEB-87F) — 100 events

Same shape as #2357 (movies / episodes pages).

## Test plan
- [ ] \`deno task client:test\` passes
- [ ] Navigate rapidly between two show pages; placeholder shows briefly, no crash

### DIFF
--- a/projects/client/src/routes/shows/[slug]/+page.svelte
+++ b/projects/client/src/routes/shows/[slug]/+page.svelte
@@ -56,7 +56,11 @@
     goToSeason($show.slug, activeSeason);
   });
 
-  const isReady = $derived(!$isLoading && !isNaN(currentSeason));
+  const hasCoreData = $derived(
+    $show != null && $intl != null && $studios != null && $crew != null &&
+      $seasons != null,
+  );
+  const isReady = $derived(!$isLoading && !isNaN(currentSeason) && hasCoreData);
 </script>
 
 <TraktPage


projects/client/src/lib/sections/lists/season/SeasonList.svelte
```diff
@@ -25,14 +25,14 @@
...
-  const hasSingleSeason = $derived(seasons.length === 1);
+  const hasSingleSeason = $derived(seasons?.length === 1);
 
   const isTabletLarge = useMedia(WellKnownMediaQuery.tabletLarge);
   const isDesktop = useMedia(WellKnownMediaQuery.desktop);
   const isLargeScreen = $derived($isTabletLarge || $isDesktop);
 
   const previousSeasons = $derived(
-    seasons.filter((s) => s.number > 0 && s.number < currentSeason),
+    seasons?.filter((s) => s.number > 0 && s.number < currentSeason) ?? [],
```
seferturan: This issue might be elsewhere if this ever caused an error. In this component, the seasons array is always defined.

projects/client/src/lib/sections/lists/season/SeasonList.svelte
```diff
@@ -25,14 +25,14 @@
 
   const title = m.list_title_seasons();
   const subtitle = $derived(seasonLabel(currentSeason));
-  const hasSingleSeason = $derived(seasons.length === 1);
+  const hasSingleSeason = $derived(seasons?.length === 1);
```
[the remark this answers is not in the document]
vladjerca: You're right — the type `Season[]` should hold, and adding `?.` here just papers over an upstream bug. Reverted the SeasonList changes; the actual fix now lives in the show route's `isReady` predicate (`$seasons != null` added) plus dropping the `!` non-null assertions on the prop pass. Same shape as #2357 for movies/episodes.

projects/client/src/lib/sections/lists/season/SeasonList.svelte
vladjerca: Agreed, that was the right read. The fix moved upstream: `/shows/[slug]/+page.svelte` now requires `$seasons != null` in its `isReady` predicate, so SeasonList keeps its non-nullable `Season[]` contract.

seferturan (approved):

PR #2357, opened 2026-05-15
fix: gate movie/episode isLoading on core data presence

vladjerca:
## Summary
- **\`useMovie.ts\`**: keep \`isLoading\` true until \`movie.data\` is present (not just until fetching stops)
- **\`useEpisode.ts\`**: same — also requires \`show.data\`
- **Pages**: replace \`media={\$movie!}\` non-null assertions with truthy guards in the \`{#if}\` clause

## Why
TanStack's \`isLoading\` only reflects fetch state — a query that resolves with \`data === undefined\` (errored or empty body) flips loading to false while the data is still missing. The pages trusted \`!\$isLoading\` and asserted every prop non-null with \`!\`. During a navigation transition or after an error the summary mounted with \`undefined\` \`media\` and crashed at \`media.slug\`.

Per review feedback, the right fix lives in the hooks: \`isLoading\` now encodes "core payload is ready," not just "the network has settled." The page-level truthy guards stay as a belt-and-suspenders and let us drop the non-null assertions for honesty.

- [TRAKT-WEB-4MS](https://trakt-tv.sentry.io/issues/TRAKT-WEB-4MS) — 1,131 events (movies)
- [TRAKT-WEB-4MR](https://trakt-tv.sentry.io/issues/TRAKT-WEB-4MR) — 4 events (episodes)

## Test plan
- [ ] \`deno task client:test\` passes
- [ ] Movie / episode pages render correctly on slow network
- [ ] Error response on movie summary: page stays in loader (instead of crashing) — confirm acceptable before merge

### DIFF
--- a/projects/client/src/routes/movies/[slug]/+page.svelte
+++ b/projects/client/src/routes/movies/[slug]/+page.svelte
@@ -24,12 +24,12 @@
     <NavbarStateSetter mode="minimal" />
   </RenderFor>
 
-  {#if !$isLoading}
+  {#if !$isLoading && $movie && $studios && $crew && $intl}
     <MovieSummary
-      media={$movie!}
-      studios={$studios!}
-      crew={$crew!}
-      intl={$intl!}
+      media={$movie}
+      studios={$studios}
+      crew={$crew}
+      intl={$intl}
       streamOn={$streamOn}
       videos={$videos}
       sentiment={$sentiment}
--- a/projects/client/src/routes/shows/[slug]/seasons/[season]/episodes/[episode]/+page.svelte
+++ b/projects/client/src/routes/shows/[slug]/seasons/[season]/episodes/[episode]/+page.svelte
@@ -30,15 +30,15 @@
     <NavbarStateSetter mode="minimal" />
   </RenderFor>
 
-  {#if !$isLoading}
+  {#if !$isLoading && $show && $showIntl && $episode && $intl && $seasons && $crew}
     <EpisodeSummary
-      show={$show!}
-      showIntl={$showIntl!}
-      episode={$episode!}
-      episodeIntl={$intl!}
-      seasons={$seasons!}
+      show={$show}
+      showIntl={$showIntl}
+      episode={$episode}
+      episodeIntl={$intl}
+      seasons={$seasons}
       streamOn={$streamOn}
-      crew={$crew!}
+      crew={$crew}
     />
   {:else}
     <!-- TODO: remove this when we have empty state, currently prevents content jumps -->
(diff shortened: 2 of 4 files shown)


projects/client/src/routes/movies/[slug]/+page.svelte
```diff
@@ -24,12 +24,12 @@
     <NavbarStateSetter mode="minimal" />
   </RenderFor>
 
-  {#if !$isLoading}
+  {#if !$isLoading && $movie && $studios && $crew && $intl}
```
Comment (seferturan) [inline projects/client/src/routes/movies/[slug]/+page.svelte:27]: Good change to at least get rid of the assertions, but isLoading should've been false if those can be undefined. Issue might be in `useMovie.ts`

projects/client/src/routes/shows/[slug]/seasons/[season]/episodes/[episode]/+page.svelte
```diff
@@ -30,15 +30,15 @@
     <NavbarStateSetter mode="minimal" />
   </RenderFor>
 
-  {#if !$isLoading}
+  {#if !$isLoading && $show && $showIntl && $episode && $intl && $seasons && $crew}
```
Comment (seferturan) [inline projects/client/src/routes/shows/[slug]/seasons/[season]/episodes/[episode]/+page.svelte:33]: Good change to at least get rid of the assertions, but isLoading should've been false if those can be undefined. Issue might be in `useEpisode.ts`

projects/client/src/routes/movies/[slug]/+page.svelte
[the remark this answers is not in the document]
Comment (vladjerca) [reply projects/client/src/routes/movies/[slug]/+page.svelte:27]: $videos is mapped to `.data ?? []` in `useMovie`, so it's always an array and never the trigger for the crash. The actual gap was that `isLoading` could go false while `$movie.data` was still undefined — addressed at the source in `useMovie.ts`/`useEpisode.ts` (see the other thread). Left the page-level truthy guards in as defence-in-depth.

Comment (vladjerca) [reply projects/client/src/routes/movies/[slug]/+page.svelte:27]: Agreed — the bug was upstream. `useMovie.ts` now keeps `isLoading=true` until `movie.data` is actually present, not just until fetching stops (TanStack flips `isFetching=false` even when data resolves to undefined on error/empty body). Page guards kept as belt-and-suspenders, but they should no longer be load-bearing.

projects/client/src/routes/shows/[slug]/seasons/[season]/episodes/[episode]/+page.svelte
Comment (vladjerca) [reply projects/client/src/routes/shows/[slug]/seasons/[season]/episodes/[episode]/+page.svelte:33]: Same fix applied to `useEpisode.ts`: `isLoading` now also requires `episode.data` and `show.data` to be present. That removes the window where `!$isLoading` was reachable with undefined core data.

seferturan (approved):

PR #2428, opened 2026-05-28
fix(history): requests history with explicit end date in paginated page

### LINKED ISSUE #2427: History does not go back far enough
When viewing the paginated history, you can't go back in your entire history.

seferturan:
## 🎶 Notes 🎶

- Fixes #2427
- Uses explicit end date for paginated history.


seferturan:
Will be fixed on API level

PR #2502, opened 2026-06-04
fix: links to app/play store

StefanFriptu:
clicking on "download from appstore/play store" were broken because:
1. redirect was applied to links
2. current links (trakt.tv/a/trakt-ios) are broken

this fixed links


seferturan (approved):
Applied a fixup here:

- For now I switched it back to the 'friendly' links.
- Instead of white listing all external links in the guard, I added an explicit whitelist. This way things like the extras URL's still won't be able to be scraped

Issue #2304, opened 2026-05-11
Add a 'status' filter


seferturan:
For example to be able to filter on things like `canceled` statuses.

PR #2510, opened 2026-06-05
feat: enable status global filter

michaldrabik:
Adds Status filter to navbar filters. Lets users narrow shows by airing status (returning, upcoming, ended, canceled, etc).

  Simple mode keeps things tidy: 4 grouped buckets that combine related statuses (e.g., "Returning Series" covers both returning series and continuing). Advanced mode exposes all 8 individual ShowStatus values for
  power users who want finer control.

  Status dropdown sits below Certification + Region in simple view. FilterGroup got a small grid tweak so any lone odd row item spans full width — keeps layout balanced no matter how many list filters exist.

  Summary

  - New Status filter (FilterKey.Status = 'statuses').
  - Simple view: 4 grouped options (returning, upcoming, ended, canceled).
  - Advanced view: 8 separate multi-select options, one per ShowStatus.
  - FilterGroup lays out a lone odd child across both columns.
  - New i18n strings: header_status, translated_value_status_*.

  Test plan

  - [ ] Open filter sidebar → Status visible below Certification/Region.
  - [ ] Status spans full width when alone in its row.
  - [ ] Simple Status pick → URL ?statuses=... matches grouped CSV.
  - [ ] Advanced Status → 8 separate multi-select entries.
  - [ ] Multi-select advanced → URL reflects combined CSV.
  - [ ] Reload with ?statuses= param → filter restores.
  - [ ] i18n labels render correctly.

### DIFF
--- a/projects/client/src/lib/features/filters/_internal/constants.ts
+++ b/projects/client/src/lib/features/filters/_internal/constants.ts
@@ -176,7 +176,39 @@ const COUNTRY_FILTER: Filter = {
   },
 };
 
+const STATUS_FILTER: Filter = {
+  label: m.header_status,
+  key: FilterKey.Status,
+  type: 'list',
+  options: [
+    { label: m.translated_value_status_released, value: 'returning series,continuing,released' },
+    { label: m.translated_value_status_upcoming, value: 'upcoming,in production,planned,post production' },
+    { label: m.translated_value_status_ended, value: 'ended' },
+    { label: m.translated_value_status_canceled, value: 'canceled' },
+    { label: m.translated_value_status_rumored, value: 'rumored' },
+  ],
+  advanced: {
+    type: 'multi-select',
+    options: [
+      {
+        label: m.translated_value_status_returning_series,
+        value: 'returning series',
+      },
+      { label: m.translated_value_status_released, value: 'released' },
+      { label: m.translated_value_status_continuing, value: 'continuing' },
+      { label: m.translated_value_status_upcoming, value: 'upcoming' },
+      { label: m.translated_value_status_in_production, value: 'in production' },
+      { label: m.translated_value_status_post_production, value: 'post production' },
+      { label: m.translated_value_status_planned, value: 'planned' },
+      { label: m.translated_value_status_ended, value: 'ended' },
+      { label: m.translated_value_status_canceled, value: 'canceled' },
+      { label: m.translated_value_status_rumored, value: 'rumored' },
+    ],
+  },
+};
+
 export const FILTERS = [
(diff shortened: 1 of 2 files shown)


michaldrabik:
Since Status is only relevant for Shows I am considering either changing the label to Show Status or/and displaying Status menu only for Media/Show

@ElMagnea @seferturan Input needed for best approach at the moment.
cc @vladjerca 

image

seferturan (changes requested):
@michaldrabik, I think it can be a filter applicable to both movies and shows. A while back @vladjerca and I had a look at it and thought this could be a mapping of label to value:

- `Upcoming` => `In Production`, `Planned`, `Upcoming`, `Post Production`
- `Released` => `Returning Series`, `Continuing`, `Released`
- `Canceled` => `Canceled`
- `Ended` => `Ended`
- `Canceled` => `Canceled`
- `Rumored` => `Rumored`

The `upcoming`, `released`, and `canceled` values would filter on both movies and shows. `Ended` and `Canceled` would work only on shows, and `rumored` would work only on movies.
In the UI we can always just show all labels and still send the values; in unsupported cases it would simply result in that `filter has no result` placeholder

michaldrabik:
I might have miss-thought that movies do not carry a status field at all. But it would make sense as described if they do! 

This PR then should wait a bit as some more updates for trakt-workers will be needed first.

michaldrabik:
@seferturan Workers are now merged and updated the Status filters field, should be now ready for re-review.

projects/client/src/lib/sections/navbar/components/filter/filters/_internal/FilterGroup.svelte
```diff
@@ -11,5 +11,9 @@
     display: grid;
     grid-template-columns: 1fr 1fr;
     gap: var(--gap-s);
+
+    & > :global(:first-child:nth-child(odd)) {
```
seferturan: Not sure I'm the biggest fan of having such wide dropdowns 😅

Image

vs.

Image

Thoughts @ElMagnea ?

seferturan (changes requested):

projects/client/src/lib/sections/navbar/components/filter/filters/_internal/FilterGroup.svelte
ElMagnea: I'll go with option B please; and some fries. Thank you!

seferturan (approved):

StefanFriptu:
closes https://github.com/trakt/trakt-web/issues/2304

PR #2562, opened 2026-06-09
Episode: fix media card links

### LINKED ISSUE #2139: Clicking on episode leads to show page instead of episode page
### Code of Conduct

Continue watching and history sections.

When clicking on an episode from the Home page section it will correctly open the specific episode, but when selecting it on the expanded view page the website will redirect to the full show page instead. 

https://github.com/user-attachments/assets/ad4e9dd9-b53c-404a-a4b8-bf823405b9c0

### Additional information

_No response_

rectifyer:
Closes #2139

# Problem

On the Home page, episode cards in **Continue Watching** and **History** correctly open the specific episode. But on the **expanded/full-page** versions of those sections, clicking an episode incorrectly navigated to the show page instead.

# Cause

The home sections render the cover-style `EpisodeCard`, whose cover always links to the episode. The expanded pages render the summary-style `MediaSummaryCard`, which only linked to the episode when an explicit `context="show"` prop was passed — otherwise it fell back to the show URL:

```ts
const href = $derived(
  isShowContext && rest.type === "episode"
    ? UrlBuilder.episode(media.slug, rest.episode.season, rest.episode.number)
    : UrlBuilder.media(media.type, media.slug), // ← show page
);
```

Neither expanded path passes `context="show"`:

- Continue Watching expanded → `variant: "next"`
- History expanded → `variant: "activity"`

…so every episode there linked to the show.

# Fix

An episode card's primary click target should always be the episode, matching `EpisodeCard`. Dropped the `isShowContext` requirement from the `href` derivation in `MediaSummaryCard.svelte`:

```ts
const href = $derived(
  rest.type === "episode"
    ? UrlBuilder.episode(media.slug, rest.episode.season, rest.episode.number)
    : UrlBuilder.media(media.type, media.slug),
);
```

The title/display logic that legitimately depends on `isShowContext` (showing the show name in mixed lists) is untouched — only the link target changes.

# Scope

`MediaSummaryCard` is the shared chokepoint for all summary-style episode cards, so this one-line change also fixes episodes in **user lists** (summary view) and **upcoming/calendar** episodes rendered in summary/compact/minimal layouts. The cover-style `EpisodeCard` was already correct everywhere.

# Testing

- `svelte-check`: 0 errors, 0 warnings
- `deno fmt`: clean
- Verified episodes in expanded Continue Watching and History now route to `/shows/{slug}/seasons/{season}/episodes/{number}`

### DIFF
--- a/projects/client/src/lib/sections/lists/components/MediaSummaryCard.svelte
+++ b/projects/client/src/lib/sections/lists/components/MediaSummaryCard.svelte
@@ -140,7 +140,7 @@
   });
 
   const href = $derived(
-    isShowContext && rest.type === "episode"
+    rest.type === "episode"
       ? UrlBuilder.episode(media.slug, rest.episode.season, rest.episode.number)
       : UrlBuilder.media(media.type, media.slug),
   );


seferturan (changes requested):
This was done on purpose at the time; e.g. to keep it more consistent with clicking the show title of the footer in the cards in `Continue watching`. When drilled down, and you hover over it, it also underlines the show title.

Though, I agree this should be changed/fixed. But I think it should be more similar to the behavior in the android app:
- Clicking the poster should take to the show.
- Clicking the rest of the card can go to the episode (on the larger screen variant, when hovering over it, the episode title should then be underlined instead of the show name).
- For movies, both the poster and rest of the card should go to the movies.

vladjerca:
> This was done on purpose at the time; e.g. to keep it more consistent with clicking the show title of the footer in the cards in `Continue watching`. When drilled down, and you hover over it, it also underlines the show title.
> 
> Though, I agree this should be changed/fixed. But I think it should be more similar to the behavior in the android app:
> 
> * Clicking the poster should take to the show.
> * Clicking the rest of the card can go to the episode (on the larger screen variant, when hovering over it, the episode title should then be underlined instead of the show name).
> * For movies, both the poster and rest of the card should go to the movies.

@seferturan I believe the real fix here is prioritizing the new episode / season views since those will render this entire conversation mute.

rectifyer:
Closing since there is another draft PR that will handle this.

PR #2678, opened 2026-06-22
feat(summary): remove VIP requirement for set cover image

Marius-TV:
Removed VIP gating from the set cover image feature, making it available to all users


seferturan (approved):
I updated the api to support this for non VIP users. Also applied a small fixup to get rid of dead code and deal with the loading state.

PR #2746, opened 2026-06-30
feat(button): adopt outline flat style as the default look

vladjerca:
## What

Makes the redesigned button look the app-wide default (builds on the bits-ui port from #2733).

- **Flat (the default style) becomes an outline** instead of a solid fill: accent ring + accent text/icon, with a primary/secondary hierarchy - primary is heavier (2px ring + a tint of fill), secondary is a quiet hairline ring.
- **Rounder radius** (`--border-radius-l`) for a more tactile feel.
- **Hover/press are pure colour shifts** - stronger fill tint + brighter ring on hover, deeper fill + full ring on press. No lift, no glow (both read as dated/cheap), no size jump.
- A swap-immune `--accent-button` / `--accent-action-button` is derived per colour so the outline's ring/text survive the existing hover fg/bg var-swap.
- Action (icon) buttons share the same outline language so they stay consistent with labelled buttons.
- Ghost, underlined and disabled states keep their baseline behaviour.

The 5 explored variants (aurora/glass/elevated/outline/tactile) and the design-system toggle were scaffolding; this PR ships only the chosen merge ("outline + tactile radius"). The DS buttons page is unchanged from main.

## Notes

- This flips every filled primary CTA to an outline app-wide. On busy/cover backgrounds a solid fill sometimes carried the contrast - worth a look during review. If a few high-emphasis spots need the old solid fill we can add a dedicated `solid` style rather than reverting the default.

## Test

- `deno task check`: 0 errors / 0 warnings
- `deno task test:unit` (buttons): 24 passed
- Verified live across home / summary / design-system routes.

### DIFF
--- a/projects/client/src/lib/components/buttons/ActionButton.svelte
+++ b/projects/client/src/lib/components/buttons/ActionButton.svelte
@@ -95,6 +95,11 @@
     $bg: var(--color-background-#{$color});
     $fg: var(--color-foreground-#{$color});
 
+    // Stable per-colour accent, immune to the hover fg/bg var-swap.
+    :global(#{$base}[data-color=#{$color}]) {
+      --accent-action-button: #{$bg};
+    }
+
     @include variant-styles($base, $color, primary, $bg, $fg);
     @include variant-styles($base, $color, secondary, $fg, $bg);
 
@@ -145,7 +150,7 @@
     align-items: center;
     flex-shrink: 0;
 
-    border-radius: var(--border-radius-m);
+    border-radius: var(--border-radius-l);
     background-color: var(--color-background-action-button);
     color: var(--color-foreground-action-button);
 
@@ -177,18 +182,42 @@
     display: none;
   }
 
+  // Flat is an outline matching the labelled buttons: accent ring + accent
+  // icon, primary heavier than secondary, colour-shift hover, no glow.
+  :global(#{$b}[data-style=flat][data-variant=primary]#{$on}) {
+    background-color: color-mix(in srgb, var(--accent-action-button) 14%, transparent);
+    color: var(--accent-action-button);
+    border: var(--border-thickness-xs) solid
+      color-mix(in srgb, var(--accent-action-button) 65%, transparent);
+  }
+
+  :global(#{$b}[data-style=flat][data-variant=secondary]#{$on}) {
+    background-color: transparent;
+    color: var(--accent-action-button);
+    border: var(--border-thickness-xxs) solid
+      color-mix(in srgb, var(--accent-action-button) 30%, transparent);
+  }
+
   @include for-mouse {
-    :global(#{$b}:hover#{$on}) {
-      box-shadow: 0 var(--ni-2) var(--ni-8) var(--ni-neg-2)
-        color-mix(
-          in srgb,
-          var(--color-background-action-button) 50%,
-          transparent
-        );
+    :global(#{$b}[data-style=flat][data-variant=primary]:hover#{$on}) {
+      background-color: color-mix(in srgb, var(--accent-action-button) 26%, transparent);
+      border-color: color-mix(in srgb, var(--accent-action-button) 95%, transparent);
+    }
+
+    :global(#{$b}[data-style=flat][data-variant=secondary]:hover#{$on}) {
+      background-color: color-mix(in srgb, var(--accent-action-button) 16%, transparent);
+      border-color: color-mix(in srgb, var(--accent-action-button) 60%, transparent);
     }
   }
 
-  :global(#{$b}:active#{$on}) {
+  :global(#{$b}[data-style=flat]:active#{$on}) {
+    transform: scale(0.92);
+    background-color: color-mix(in srgb, var(--accent-action-button) 36%, transparent);
+    border-color: var(--accent-action-button);
+    box-shadow: none;
+  }
+
+  :global(#{$b}[data-style=ghost]:active#{$on}) {
     transform: scale(0.92);
     box-shadow: none;
   }
@@ -206,28 +235,6 @@
   :global(#{$b}[data-size=large]) {
     scale: 1.2;
     margin: var(--ni-4);
-  }
-
-  :global(#{$b}[data-variant=secondary]:not([data-style=ghost])#{$on}) {
-    background-color: color-mix(
-      in srgb,
-      var(--color-foreground-action-button) 5%,
-      transparent
-    );
-    border: var(--border-thickness-xxs) solid
-      color-mix(
-        in srgb,
-        var(--color-foreground-action-button) 50%,
-        transparent
-      );
-    color: var(--color-text-primary);
-  }
-
-  @include for-mouse {
-    :global(#{$b}[data-variant=secondary]:not([data-style=ghost])#{$on}:hover) {
-      background-color: var(--color-background-action-button);
-      color: var(--color-foreground-action-button);
-    }
   }
 
   :global(#{$b}[data-style=ghost]) {
--- a/projects/client/src/lib/components/buttons/ActionButton.svelte
+++ b/projects/client/src/lib/components/buttons/ActionButton.svelte
@@ -106,6 +106,11 @@
     $bg: var(--color-background-#{$color});
     $fg: var(--color-foreground-#{$color});
 
+    :global(#{$base}[data-color=#{$color}]) {
+      --accent-action-button: var(--color-accent-#{$color}, #{$bg});
+      --surface-action-button: #{$bg};
+    }
+
     @include variant-styles($base, $color, primary, $bg, $fg);
     @include variant-styles($base, $color, secondary, $fg, $bg);
 
@@ -188,18 +193,14 @@
     display: none;
   }
 
-  @include for-mouse {
-    :global(#{$b}:hover#{$on}) {
-      box-shadow: 0 var(--ni-2) var(--ni-8) var(--ni-neg-2)
-        color-mix(
-          in srgb,
-          var(--color-background-action-button) 50%,
-          transparent
-        );
-    }
-  }
+  @include flat-outline-button(
+    $b,
+    var(--accent-action-button),
+    var(--surface-action-button),
+    scale(0.92)
+  );
 
-  :global(#{$b}:active#{$on}) {
+  :global(#{$b}[data-style=ghost]:active#{$on}) {
     transform: scale(0.92);
     box-shadow: none;
   }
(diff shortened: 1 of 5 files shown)


projects/client/src/lib/components/buttons/Button.svelte
```diff
@@ -344,16 +350,39 @@
...
 
+  // Flat is an outline: no heavy fill, accent ring + accent text. Primary
+  // leans heavier (thicker ring + a tint of fill); secondary is a quiet
+  // hairline ring. Hover/press are pure colour shifts - no lift, no glow.
+  :global(#{$b}[data-style=flat][data-variant=primary]#{$on}) {
+    background: color-mix(in srgb, var(--accent-button) 14%, transparent);
+    color: var(--accent-button);
+    border: var(--border-thickness-xs) solid
+      color-mix(in srgb, var(--accent-button) 65%, transparent);
+  }
```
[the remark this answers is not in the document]
seferturan: Border's gone now, swapped it for an inset `box-shadow`.

The real border was affecting layout: `flex: 1` in a column container defeats `height`, so the border width ended up in the content-based min height and enabled buttons came out a few px taller than disabled ones.

`box-shadow` was already in `transition-property`, so the ring transitions for free 👌

projects/client/src/lib/components/buttons/ActionButton.svelte
```diff
@@ -177,18 +182,42 @@
...
   }
 
+  // Flat is an outline matching the labelled buttons: accent ring + accent
+  // icon, primary heavier than secondary, colour-shift hover, no glow.
+  :global(#{$b}[data-style=flat][data-variant=primary]#{$on}) {
+    background-color: color-mix(in srgb, var(--accent-action-button) 14%, transparent);
+    color: var(--accent-action-button);
+    border: var(--border-thickness-xs) solid
+      color-mix(in srgb, var(--accent-action-button) 65%, transparent);
+  }
```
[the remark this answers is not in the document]
seferturan: Same here, border is gone in favour of an inset `box-shadow`.

Keeps the ring out of layout so disabled and enabled action buttons stay the same size, and `box-shadow` is already transitioned 👌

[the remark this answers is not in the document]
seferturan: Same as the earlier thread on this line 👌

projects/client/src/lib/components/buttons/Button.svelte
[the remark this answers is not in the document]
seferturan: Same as the earlier thread on this line, the border is gone 👌

projects/client/src/lib/components/buttons/Button.svelte
```diff
@@ -251,7 +257,7 @@
   :global(#{$b}:active[disabled]) {
     height: var(--button-height);
     box-sizing: border-box;
-    border-radius: var(--border-radius-m);
+    border-radius: var(--border-radius-l);
```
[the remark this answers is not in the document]
seferturan: Went the other way on this one. The base is back on `--border-radius-m`, `-l` no longer matched the radius on the selects (`SelectBase`).

So small scaling off `-m` is correct as is.

projects/client/src/lib/components/buttons/ActionButton.svelte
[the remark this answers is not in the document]
seferturan: Dupe of the earlier passes, handled 👌

projects/client/src/lib/components/buttons/Button.svelte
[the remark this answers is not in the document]
seferturan: Dupe of the earlier passes, handled 👌

seferturan:
Added a fixup:

- Got rid of some code duplication
- Border radius matches the one from the `Select` component
- Outline is now done using a box shadow to avoid layout jumps between disabled/enabled state
- Applied the custom style for the button on the landing page
- Updated the colors to improve contrast ratio with the background of the button. These should now be compatible with the WCAG (https://www.w3.org/WAI/WCAG22/Understanding/contrast-minimum.html)

@rectifyer @rudf0rd, on my screen these now look fine contrast wise. But I'd like an extra check on this, please see if these look ok on your screens. Mostly in these places: changing account details settings, data settings, creating lists, adding reviews, etc.

As a follow-up, I think we can simplify these components a bit more, and settle on a single design instead of having this and the outlined version. cc @ElMagnea

projects/client/src/style/scss/mixins/index.scss
```diff
@@ -245,6 +245,54 @@
...
+      box-shadow: inset 0 0 0 var(--flat-ring-width)
+        color-mix(in srgb, #{$accent} 60%, transparent);
+    }
+  }
+
+  :global(#{$selector}[data-style='flat']:active#{$on}) {
+    transform: $active-transform;
+    background-color: color-mix(in srgb, #{$surface} 36%, transparent);
+    box-shadow: inset 0 0 0 var(--flat-ring-width) #{$accent};
+  }
```
rectifyer: The hover rules above outrank this one: `[data-style][data-variant]:hover` carries one more attribute selector than `[data-style]:active`, and a mouse press always also matches `:hover`. So on desktop the press state's deeper fill and full ring never render (hover's `background-color`/`box-shadow` win) and only the transform applies. Touch is unaffected, which makes this easy to miss.

Emitting the press rule per variant makes it tie with hover on specificity and win on source order:

```suggestion
  :global(#{$selector}[data-style='flat'][data-variant='primary']:active#{$on}),
  :global(#{$selector}[data-style='flat'][data-variant='secondary']:active#{$on}) {
    transform: $active-transform;
    background-color: color-mix(in srgb, #{$surface} 36%, transparent);
    box-shadow: inset 0 0 0 var(--flat-ring-width) #{$accent};
  }
```

projects/client/src/lib/components/buttons/Button.svelte
```diff
@@ -347,18 +352,13 @@
...
-    transform: scale(calc(var(--scale-factor-button) * 0.97));
-    box-shadow: none;
-  }
+  @include flat-outline-button(
+    $b,
+    var(--accent-button),
+    var(--surface-button),
+    scale(calc(var(--scale-factor-button) * 0.97)),
+    ":not([data-color='custom'])"
+  );
```
rectifyer, line 363: Excluding `custom` to keep the solid fill makes sense, but the old `flat:hover` (lift) and `flat:active` (scale) rules are deleted, so `custom` flat buttons are left with no press feedback at all. Since `custom` is the component's default `color`, this hits real surfaces: the ConfirmationDialog destructive action, UpgradeButton, the error pages, JoinForFreeButton, and several other call sites now only swap colours on hover and don't react to press.

Worth keeping at least the press transform for the excluded colour:

```suggestion
  @include flat-outline-button(
    $b,
    var(--accent-button),
    var(--surface-button),
    scale(calc(var(--scale-factor-button) * 0.97)),
    ":not([data-color='custom'])"
  );

  // `custom` keeps its solid fill but should still react to presses now that
  // the old flat active rule is gone.
  :global(#{$b}[data-style=flat][data-color=custom]:active#{$on}) {
    transform: scale(calc(var(--scale-factor-button) * 0.97));
  }
```

projects/client/src/lib/components/buttons/Button.svelte
```diff
@@ -132,6 +132,11 @@
     $bg: var(--color-background-#{$color});
     $fg: var(--color-foreground-#{$color});
 
+    :global(#{$base}[data-color=#{$color}]) {
+      --accent-button: var(--color-accent-#{$color}, #{$bg});
+      --surface-button: #{$bg};
+    }
```
rectifyer: Nit: for `custom` this emits `--accent-button: var(--color-accent-custom, ...)`, but `--color-accent-custom` is never defined in `modes.scss` and the mixin below excludes `data-color=custom` anyway, so the var is dead for that colour. Skipping it keeps the intent (custom stays solid) obvious:

```suggestion
    @if $color != "custom" {
      :global(#{$base}[data-color=#{$color}]) {
        --accent-button: var(--color-accent-#{$color}, #{$bg});
        --surface-button: #{$bg};
      }
    }
```

projects/client/src/style/scss/mixins/index.scss
```diff
@@ -245,6 +245,54 @@
...
+@mixin flat-outline-button(
+  $selector,
+  $accent,
+  $surface,
+  $active-transform,
+  $exclude: ""
+) {
+  $on: ":not([disabled]):not([aria-disabled='true'])#{$exclude}";
+
+  :global(#{$selector}[data-style='flat'][data-variant='primary']#{$on}) {
```
rectifyer, line 257: Nit: the button components write these selectors unquoted (`[data-style=flat]`), the mixin quotes them (`[data-style='flat']`). Both compile to the same CSS, but since the mixin's selectors have to stay in sync with the component ones, aligning on one form would help grep-ability.

rectifyer (changes requested):
Love the direction, and the execution is mostly right: the swap-immune accent vars are the correct fix for the hover fg/bg var-swap, the accent tokens live in both theme mixins (and track the seasonal palettes for free since they reference the purple ramp), and extracting the shared mixin instead of duplicating it across both buttons is the right call.

Requesting changes for two behaviour issues I hit while testing the states:

- **Press-state colours are unreachable with a mouse.** The mixin's hover rules outrank its single `:active` rule, and a mouse press always also hovers, so the "deeper fill + full ring" press only ever renders on touch. Inline comment with a fix.
- **`color="custom"` flat buttons (the default colour) lose all press feedback.** The old flat hover/active rules were removed and the mixin excludes `custom`, leaving those buttons inert on press. Inline comment.

One note on the PR description: it mentions the rounder `--border-radius-l` radius, but the diff contains no radius change (buttons still use `--border-radius-m`). Leftover from the explored variants?

Not requesting a change, but worth an eyeball before merge per your own note: the secondary hairline ring at 30% accent is quite subtle over busy cover art, especially in light mode.

projects/client/src/style/scss/mixins/index.scss
seferturan: Good catch 👌

Was staring right past it. Hover is `(0,5,0)` with the extra `[data-variant]`, the press rule was `(0,4,0)`, and a mouse press always matches `:hover` too, so only the transform was landing.

Emitting it per variant now, ties with hover and wins on source order.

projects/client/src/lib/components/buttons/Button.svelte
seferturan: Yep, and worse than it looks: `custom` is the default `color` on the component, so this killed press feedback on pretty much every button in the app 😅

Restored the transform for it. Kept it as its own rule rather than passing it into the mixin, since `custom` only needs the transform and none of the fill/ring.

projects/client/src/lib/components/buttons/Button.svelte
seferturan: Done 👌

Note the same pair is live in `ActionButton`, that one does not exclude `custom` from the outline, so only doing it here.

projects/client/src/style/scss/mixins/index.scss
seferturan: Leaving this one for now, the churn is not really worth it in the middle of the PR.

Agree in principle though. Will do a pass over the whole mixins file separately.

seferturan (approved):

PR #2753, opened 2026-07-01
YIR: auto redirect link

rectifyer:
# Add /yir short URL for year in review

Adds a short, memorable URL for a user's year in review that mirrors the existing `/users/:user/mir` month-in-review short URL.

Visiting `/users/:user/yir` (e.g. `/users/me/yir`, `/users/justin/yir`) now redirects to that user's full year-in-review page. The year is resolved automatically so the link never needs updating as the calendar rolls over.

# Behavior

- Redirects `/users/:user/yir` → `/users/:user/year/:year`.
- Defaults to the current year, but falls back to the previous year during January and February, since the current year has barely any data that early.
- Preserves query params (e.g. `?embedded_mode&slurm=1`) across the redirect.
- Uses a `307` temporary redirect so the year rollover is never cached as permanent.

# Notes

This complements the existing `/users/:user/mir` route, which already redirects to the previous month's review — no changes were needed there.

### DIFF
--- a/projects/client/src/routes/users/[user]/yir/+page.ts
+++ b/projects/client/src/routes/users/[user]/yir/+page.ts
@@ -1,0 +1,16 @@
+import { getYearInReviewYear } from '$lib/utils/date/getYearInReviewYear.ts';
+import { UrlBuilder } from '$lib/utils/url/UrlBuilder.ts';
+import { redirect } from '@sveltejs/kit';
+import type { PageLoad } from './$types';
+
+// Short URL: /users/:user/yir redirects to that user's year in review.
+// During January and February the previous year is used, since the current
+// year has barely any data yet.
+// Query params (e.g. ?embedded_mode&slurm=1) are preserved across the redirect.
+export const load: PageLoad = ({ params, url }) => {
+  const year = getYearInReviewYear(new Date());
+
+  const target = UrlBuilder.users(params.user).yearToDate(year);
+
+  return redirect(307, `${target}${url.search}`);
+};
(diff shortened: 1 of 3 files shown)


projects/client/src/routes/users/[user]/yir/+page.ts
```diff
@@ -0,0 +1,17 @@
...
+// Query params (e.g. ?embedded_mode&slurm=1) are preserved across the redirect.
+export const load: PageLoad = ({ params, url }) => {
+  const now = new Date();
+  const isEarlyYear = now.getMonth() <= 1; // January (0) or February (1)
+  const year = isEarlyYear ? now.getFullYear() - 1 : now.getFullYear();
+
+  const target = UrlBuilder.users(params.user).yearToDate(year);
+
+  return redirect(307, `${target}${url.search}`);
+};
```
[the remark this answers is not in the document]
Comment (rectifyer) [reply projects/client/src/routes/users/[user]/yir/+page.ts:16]: @gemini-code-assist code update, please check now

projects/client/src/routes/users/[user]/yir/+page.ts
```diff
@@ -0,0 +1,16 @@
+import { getYearInReviewYear } from '$lib/utils/date/getYearInReviewYear.ts';
+import { UrlBuilder } from '$lib/utils/url/UrlBuilder.ts';
+import { redirect } from '@sveltejs/kit';
+import type { PageLoad } from './$types';
+
+// Short URL: /users/:user/yir redirects to that user's year in review.
```
Comment (seferturan) [inline projects/client/src/routes/users/[user]/yir/+page.ts:6]: Shouldn't these also support `/users/:user/yir/:year` optionally? e.g. only use `getYearInReviewYear` if there is no year param?

seferturan (changes requested):

projects/client/src/routes/users/[user]/yir/+page.ts
Comment (rectifyer) [reply projects/client/src/routes/users/[user]/yir/+page.ts:6]: The original `/users/:user/yir/:year` route still works, this just adds the new short link without a year.

seferturan (approved):

PR #2774, opened 2026-07-02
feat(comments): open direct comment links in a drawer

kevincador:
## Summary
Adds support for `/comments/:id` links by rendering the home page with a direct comment drawer that fetches and displays the requested comment.

## Changes
- Extracts the home page into a reusable `HomePage` section.
- Adds `/comments/[id]` route that opens a drawer over the home page.
- Adds a direct comment drawer/card backed by the existing `commentQuery`.

## Screenshot

eg: /comments/327791

Screenshot 2026-07-02 at 13 50 28


seferturan (changes requested):
I don't think we should display this on the home page like this. The comment is way out of context.

We should be able to detect what media the comment is on, and display it on the correct summary page.

v0 can maybe use this drawer you have on the correct show/movie summary page.


v0+: ideally, even more in context:

- In case of movie/show, the full reviews drawer, and scroll to the comment. It sort of already does that, but not for id's that fall out of the initial query limit 😅
- In case of seasons/episodes, maybe even in the season/episode drawer itself.

seferturan (approved):

PR #2810, opened 2026-07-06
feat(settings): add CSV to raw export option

kevincador:
## Summary

- Adds JSON/CSV choices to the raw export button on data settings.
- Converts raw API response payloads to CSV files before creating the export ZIP.
- Keeps empty `[]` and `{}` responses as empty CSV files.
- Tracks the selected export format in data export analytics.

## Screenshots

Screenshot 2026-07-06 at 11 01 47

### DIFF
--- a/projects/client/i18n/meta/en.json
+++ b/projects/client/i18n/meta/en.json
@@ -4788,6 +4788,14 @@
       "default": "Export a ZIP file containing your raw data.",
       "description": "Aria-label for the button that allows users to export their raw data."
     },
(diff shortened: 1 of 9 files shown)


vladjerca:
Thanks Kevin. Leaning towards not merging this - exporting is about handing users their raw data, not reshaping it. CSV of our nested payloads ends up half JSON-in-cells anyway (arrays get stringified), so we'd be owning a lossy transform that isn't really our job. Raw JSON export already covers portability; format conversion is downstream of us.

PR #2894, opened 2026-07-13
fix(search): use media.trakt.tv for storage host

StefanFriptu:
Search poster/headshot URLs were pinned to `walter-r2.trakt.tv`. `walter-r2` should be gone — `media.trakt.tv` is the storage host now.

Now that Typesense returns fully-formed `media.trakt.tv` URLs, the old hardcoded host double-prepended (e.g. `https://walter-r2.trakt.tv/https://media.trakt.tv/…jpg.webp.webp`).

- Storage host → `media.trakt.tv`
- Pass through URLs that already carry a host (no double host)
- Normalize the leading slash whether the path comes with or without it
- Don't duplicate the `.webp` extension

### DIFF
--- a/projects/client/src/lib/requests/queries/search/response/prependStorageHost.ts
+++ b/projects/client/src/lib/requests/queries/search/response/prependStorageHost.ts
@@ -1,4 +1,4 @@
-const WALTER = 'walter-r2.trakt.tv';
+const STORAGE_HOST = 'media.trakt.tv';
 
 export function prependStorageHost(
   url: string | Nil,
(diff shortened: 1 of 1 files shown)


projects/client/src/lib/requests/queries/search/response/prependStorageHost.ts
```diff
@@ -8,7 +8,11 @@ export function prependStorageHost(
     return '';
   }
 
-  const imageUrl = url.startsWith(WALTER) ? url as HttpsUrl : `${WALTER}${url}`;
+  const hasHost = url.startsWith('http') || url.includes('.trakt.tv');
+  const path = url.startsWith('/') ? url : `/${url}`;
+  const imageUrl = hasHost ? url : `${STORAGE_HOST}${path}`;
```
vladjerca: Too much logic @StefanFriptu we know what we need to prepend. This branching doesn't exist and should never exist.

vladjerca (changes requested):

projects/client/src/lib/requests/queries/search/response/prependStorageHost.ts
vladjerca: Or is the branching required to preserve multi-env support?

StefanFriptu: i agree, it's not needed - I wanted to make it more fail-safe (i.e. if by accident we put a poster with the whole `https://media.trakt.tv`, this will handle it). Posters from old typesense instance had `/` before the actual path. I'll keep just that change - check whether poster path has it or not

vladjerca (approved):

PR #2911, opened 2026-07-15
feat: streak cells UI tweak

michaldrabik:
- a proposal PR (well actually Android does this at the moment :P)
- currently both busted past days and yet to come days share the same grey fill which could be confusing
- PR proposes an outline for future days (and today without activity -> a buckets yet to be filled or busted) as differentiation from "busted" days that already passed

cc @tysonkerridge @kevincador @ElMagnea @vladjerca for syncs

image


vladjerca (approved):
Yeah I like it 👌

PR #2913, opened 2026-07-15
feat(recommended): cap free-user dashboard recs with VIP upsell card

ElMagnea:
## What

Free users now get a taste of recommendations on the dashboard — the first **3 items** — followed by a **VIP upsell card** in the 4th slot. VIP users are unaffected.

## Design

Screenshot 2026-07-15 at 20 00 41

- Portrait card sized to match the recommendation posters.
- Darkened brand gradient (purple → blue → red, using each family's darker steps so white text stays legible).
- Stroked-out, uppercase **GET VIP** button (transparent fill, white stroke).

## How it works

- **Scope:** dashboard `RecommendedList` only. The `/discover/recommended` paginated page is untouched.
- **Cap:** free users are detected via `useUser` (`$user != null && !$user.isVip`) and the list is sliced client-side to 3. No query/backend change, so nothing to refetch if a user upgrades.
- **4th slot:** `SectionList` already renders its trailing `ctaItem` only when `items.length <= 4`. Capping to 3 makes the upsell render as the 4th card for free users automatically; VIP users load the full 10 (>4) so the slot never appears for them.
- **Analytics:** the button fires `AnalyticsEvent.VipUpsell` with source `recommended-list`.
- **i18n:** new key `text_vip_upsell_smart_recommendations`.

## Notes for reviewer

- **`?preview-free` dev toggle** — `RecommendedList` includes an `IS_DEV`-gated override so the free-tier layout can be previewed while signed in as any user (e.g. a VIP): visit the dashboard with `?preview-free`. It's inert in production. Happy to strip it if you'd rather keep the diff pure.
- **Contrast** — the darkened gradient was chosen so the white label on the stroked button clears ~4.5:1.

## Testing

- `deno task check` — 0 errors.
- `deno fmt` + eslint clean.
- Previewed locally via `?preview-free`.

### DIFF
--- a/projects/client/i18n/meta/en.json
+++ b/projects/client/i18n/meta/en.json
@@ -5395,6 +5395,10 @@
       "default": "Upgrade to VIP to unlock more insights.",
       "description": "Text shown to encourage non-VIP users to subscribe to explore a more detailed sentiment analysis of a movie or show."
     },
+    "text_vip_upsell_smart_recommendations": {
+      "default": "Get Smart Recs. Built on your stats, not the crowd.",
+      "description": "Heading on the upsell card shown to free users in place of the fourth recommendation on the dashboard, encouraging them to subscribe to VIP for personalized smart recommendations."
+    },
     "text_sentiment_example": {
       "default": "Example from \"{title}\"",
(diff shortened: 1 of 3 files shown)


vladjerca:
This is not true.

PR #2933, opened 2026-07-16
feat: Arabic localization and RTL support

3-pr:
This PR implements Arabic localization and RTL support for trakt-web.

### Changes
- **RTL Logical Properties**: Updated components to use CSS logical properties (e.g. `padding-inline`, `inset-inline`) ensuring the layout naturally flips for RTL scripts like Arabic.
- **Bi-directional Text**: Utilized `<bdi>` tags for metrics and text that may combine Latin and Arabic text incorrectly.
- **Locale Mapping**: Mapped the backend `ar-SA` requirement transparently so the frontend treats it as the generic `ar` locale to simplify selection.
- **Overrides**: Forced Gregorian calendar and Latin numerals (ar-u-ca-gregory-nu-latn) so that the UI does not change numerals unexpectedly when Arabic is selected.
- **Typography**: Added the Google 'Tajawal' font for a seamless visual experience for Arabic characters.


seferturan (changes requested):
Thanks for a great first contribution :D The `<bdi>` isolation changes are the right idea, let's keep them as-is for now. I'll fold them into a shared helper in a follow-up so we're not wrapping per call site, so no need to touch those further here.

As for the other things:

*** 
Every locale except en is language + region (`fa-IR`, `nl-NL`, ...), and the backend needs the region too, which is exactly why useSettings maps `ar` -> `ar-SA` and `currentUserSettingsQuery` maps it back. Naming the locale bare `ar` is what forced most of the plumbing in this PR. If we call it `ar-SA` end to end, all of this can go:

- `splitLanguageTagRaw` + `canonicalLocaleMap` (the whole ar -> ar-ar -> ar dance)
- the `sanitizeLocale` canonical fallback branch
- the `splitLanguageTag` region fallback tweak
- the `ar-*` -> `ar` map in `currentUserSettingsQuery`
- the `ar` -> `ar-SA` map in useSettings

Then it's just: rename `ar.json` -> `ar-SA.json`, and `ar` -> `ar-SA` in `settings.json` and `RTL_LOCALES`. Much smaller diff, and it keeps our "only en is region-less" invariant intact.

***
Happy to add Tajawal, would also improve things for `fa-IR` users, but the weight list needs a trim. It currently pulls 300/900 which we don't use, and is missing 600, which is our most common weight. Can we go with 400;500;600;700?

***
Two things bundled in getIntlLocale. The ca-gregory part is a real fix, please keep it, ar-SA otherwise defaults to the Hijri calendar which we don't want for air dates. The nu-latn part I'd lean towards dropping so Arabic uses its native digits like fa-IR does (fa already ships the same native-number + Latin-episode-code mix, so we'd be consistent). If the Latin digits were a deliberate call for your audience, that's fair, but then let's apply the same to fa-IR so the two RTL locales don't diverge silently.

***
You've added a bespoke `scaleX(-1)` under `[dir="rtl"]`. We already have a `trakt-icon-directional` convention for mirroring (see `components.md`). Can we route it through that, or if the star shouldn't really flip, drop it?

***
What's the `if (!locale) return ''` guard in `toMeasurement` fixing? With the new 'en' default it shouldn't be needed right?

3-pr:
> Thanks for a great first contribution :D The `<bdi>` isolation changes are the right idea, let's keep them as-is for now. I'll fold them into a shared helper in a follow-up so we're not wrapping per call site, so no need to touch those further here.
> 
> As for the other things:
> 
> Every locale except en is language + region (`fa-IR`, `nl-NL`, ...), and the backend needs the region too, which is exactly why useSettings maps `ar` -> `ar-SA` and `currentUserSettingsQuery` maps it back. Naming the locale bare `ar` is what forced most of the plumbing in this PR. If we call it `ar-SA` end to end, all of this can go:
> 
> * `splitLanguageTagRaw` + `canonicalLocaleMap` (the whole ar -> ar-ar -> ar dance)
> * the `sanitizeLocale` canonical fallback branch
> * the `splitLanguageTag` region fallback tweak
> * the `ar-*` -> `ar` map in `currentUserSettingsQuery`
> * the `ar` -> `ar-SA` map in useSettings
> 
> Then it's just: rename `ar.json` -> `ar-SA.json`, and `ar` -> `ar-SA` in `settings.json` and `RTL_LOCALES`. Much smaller diff, and it keeps our "only en is region-less" invariant intact.
> 
> Happy to add Tajawal, would also improve things for `fa-IR` users, but the weight list needs a trim. It currently pulls 300/900 which we don't use, and is missing 600, which is our most common weight. Can we go with 400;500;600;700?
> 
> Two things bundled in getIntlLocale. The ca-gregory part is a real fix, please keep it, ar-SA otherwise defaults to the Hijri calendar which we don't want for air dates. The nu-latn part I'd lean towards dropping so Arabic uses its native digits like fa-IR does (fa already ships the same native-number + Latin-episode-code mix, so we'd be consistent). If the Latin digits were a deliberate call for your audience, that's fair, but then let's apply the same to fa-IR so the two RTL locales don't diverge silently.
> 
> You've added a bespoke `scaleX(-1)` under `[dir="rtl"]`. We already have a `trakt-icon-directional` convention for mirroring (see `components.md`). Can we route it through that, or if the star shouldn't really flip, drop it?
> 
> What's the `if (!locale) return ''` guard in `toMeasurement` fixing? With the new 'en' default it shouldn't be needed right?


hey @seferturan 
Thanks so much for the detailed review — this was a great roadmap to tighten things up. Here's everything addressed:

### `<bdi>` isolation
Left exactly as-is, per your note — looking forward to the shared-helper follow-up so we're not wrapping per call site. 👍

### `ar-SA` end to end
Done, and it collapsed the diff nicely. Since the locale is now `ar-SA` throughout, I removed all the plumbing that bare `ar` had forced:
- `splitLanguageTagRaw` + `canonicalLocaleMap` (the whole `ar -> ar-ar -> ar` dance)
- the `sanitizeLocale` canonical fallback branch
- the `splitLanguageTag` region fallback tweak
- the `ar-*` -> `ar` map in `currentUserSettingsQuery`
- the `ar` -> `ar-SA` map in `useSettings`

Then just the renames: `ar.json` -> `ar-SA.json`, and `ar` -> `ar-SA` in `settings.json` and `RTL_LOCALES`. The "only `en` is region-less" invariant is intact.

### Tajawal weights
Trimmed to `400;500;600;700` (dropped 300/900, added the missing 600).

### Calendar (`ca-gregory`)
Kept — agreed, defaulting to Hijri for air dates would be wrong.

### Numerals (`nu-latn`)
I'd like to keep Latin digits for `ar-SA`, and I'd prefer to leave `fa-IR` untouched rather than align them. The reasoning: Western/Arabic numerals are the everyday norm in Saudi/Gulf digital products — ratings, runtimes, years, episode codes all read more naturally in Latin digits for this audience, whereas Persian genuinely reads better with its native numerals. So this isn't a silent divergence between the two RTL locales — it's a deliberate per-audience choice, and I've documented it in a comment next to the locale config so the intent is clear. The date path stays consistent too, since date-fns `ar-SA` already emits Latin digits. Totally open to revisiting if you feel strongly, but that's my recommendation for the Arabic UX.

### Star `scaleX(-1)`
Agreed the star glyph itself shouldn't flip — I dropped the bespoke whole-`svg` mirror entirely. What remains is scoped to the fill overlay only (`.trakt-star-fill`), to correct the partial-fill *direction* in RTL: without it the fill always grows from the left edge, so a 3.5/5 rating fills the wrong side. Since that's a fill-clip direction rather than mirroring a directional icon, `trakt-icon-directional` didn't map cleanly onto it — but if you'd prefer I route it through the convention (or handle the fill direction a different way), just say the word and I'll adjust.

### `toMeasurement` guard
Removed — it was leftover and, as you said, unnecessary with the `en` default.

Let me know if anything else needs tweaking. Thanks again for the thoughtful review and the warm welcome! 🙏

projects/client/src/lib/features/i18n/index.ts
```diff
@@ -40,7 +40,7 @@ function splitLanguageTag(languageTag: Locale): {
     'Language code is required.',
   ) as AvailableLanguage;
   const region = assertDefined(
-    language === 'en' ? DEFAULT_REGION_EN : parts.at(1),
+    parts.at(1) ?? (language === 'en' ? DEFAULT_REGION_EN : language),
```
seferturan: This can now be:
````
const region = assertDefined(
    language === 'en' ? DEFAULT_REGION_EN : parts.at(1),
    'Region code is required.',
  ) as AvailableRegion;
````
right?

File: projects/client/src/lib/sections/profile/components/_internal/ActivityCommentItem.svelte [hunk 2]
```diff
@@ -64,10 +64,14 @@
     entry.type === "episode"
       ? entry.media.title +
           " - " +
-          episodeNumberLabel({
+          // U+2066 (LRI) … U+2069 (PDI) isolate the LTR episode code so it
+          // renders correctly when embedded in an RTL paragraph. <bdi> cannot
+          // be used here because this value is interpolated via {mediaTitle}
+          // and Svelte would escape the angle brackets as literal text.
+          `\u2066${episodeNumberLabel({
```
seferturan: The `\u2066`/`\u2069` isolates trip Svelte's `bidirectional_control_characters`
  (https://svelte.dev/e/bidirectional_control_characters) warning. The escape doesn't dodge it either,
  Svelte checks the decoded string value, not the raw source, so it still sees the control char and warns.

  Since it's needed in both `ActivityCommentItem` and `Yir2024MediaCard`, let's pull it into a small helper and
  build the chars at runtime so nothing bidi ends up in the source:

````
  // lib/utils/string/isolateLtr.ts
  const LRI = String.fromCharCode(0x2066); // Left-to-Right Isolate
  const PDI = String.fromCharCode(0x2069); // Pop Directional Isolate

  /** Wrap LTR text in bidi isolates so it renders correctly inside an RTL run. */
  export const isolateLtr = (text: string): string => `${LRI}${text}${PDI}`;
````

  Then the call sites become:
````
  // ActivityCommentItem.svelte
  entry.media.title + " - " + isolateLtr(episodeNumberLabel({
    seasonNumber: entry.episode.season,
    episodeNumber: entry.episode.number,
  }))
````

````
  // Yir2024MediaCard.svelte
  isolateLtr(episodeNumberLabel({
    seasonNumber: item.episode.season,
    episodeNumber: item.episode.number,
  }))
````

  Same runtime behaviour, warning goes away 👌

seferturan (changes requested):
@3-pr thanks for the detailed response, I agree with your proposals :D Just these small minor things.

3-pr:
Both make sense, thanks! 🙏 @seferturan 

1. `i18n/index.ts` — yes, exactly. Simplified `assertDefined` to the plain `'Region code is required.'` message and kept the `as AvailableRegion` cast.
2. bidi isolates — good call. Extracted `isolateLtr` into `lib/utils/string/isolateLtr.ts`, building the chars at runtime with `String.fromCharCode`, and updated both `ActivityCommentItem.svelte` and `Yir2024MediaCard.svelte` to use it. The `bidirectional_control_characters` warning is gone and there are no bidi chars left in the source. Also removed the now-stale `<bdi>` comment.

Pushed in the latest commit.

3-pr:
@seferturan Sorry for the extra round of CI approvals here! 🙏

The last force-push was purely a commit-message reword — I renamed the
`fix(typography): …` commit to `fix(i18n): …` so it satisfies the commitlint
scope-enum. No code changed in that push.

Both of your review comments are addressed in the latest commit (84e672f):
- the region `assertDefined` now takes a plain `'Region code is required.'`
  message with the `AvailableRegion` cast, and
- the bidi isolates are extracted into a small runtime `isolateLtr` helper,
  now used by both `ActivityCommentItem` and `Yir2024MediaCard`.

Whenever you get a chance, could you re-approve the workflow run so the PR Gate
can execute again? Thanks a lot for your patience and the reviews 🙏

seferturan:
> @seferturan Sorry for the extra round of CI approvals here! 🙏
> 
> The last force-push was purely a commit-message reword — I renamed the `fix(typography): …` commit to `fix(i18n): …` so it satisfies the commitlint scope-enum. No code changed in that push.
> 
> Both of your review comments are addressed in the latest commit ([84e672f](https://github.com/trakt/trakt-web/commit/84e672f2d1c230261e8fa0b9856a4e838c3d1e4a)):
> 
> * the region `assertDefined` now takes a plain `'Region code is required.'`
>   message with the `AvailableRegion` cast, and
> * the bidi isolates are extracted into a small runtime `isolateLtr` helper,
>   now used by both `ActivityCommentItem` and `Yir2024MediaCard`.
> 
> Whenever you get a chance, could you re-approve the workflow run so the PR Gate can execute again? Thanks a lot for your patience and the reviews 🙏

Not a problem at all :D I'll also do a quick local check, and squash and merge 👌 
And thanks for this contribution 🎉

seferturan (approved):

Issue #2941, opened 2026-07-17
"Hide Watched" (App) and "Display Watched" (web) Toggle doesn't work


kromiumman:
### Code of Conduct

- [x] I have read and agree to the project's [Code of Conduct](https://github.com/trakt/trakt-web/blob/main/.github/CODE_OF_CONDUCT.md)

### Where did you encounter the bug?

Both desktop and iOS app versions.

### Please describe the bug below

When I'm viewing my calendar in both the app and on the desktop version, the "hide watched" (app) and the "display watched" (desktop) toggle do absolutely nothing.  It makes it incredibly difficult to see what I have left to watch, especially when the difference is a tiny, tiny faded white checkmark to indicate watched status.

PR #2944, opened 2026-07-17
fix(person): honor Display Watched / Watchlisted on credits pages

### LINKED ISSUE #2943: Display Watched switch not working on (at least) actor's movie page
### Code of Conduct

On Billy Bob Thornton's movies page. https://app.trakt.tv/people/billy-bob-thornton/movies?mode=media

The filter switch for Display Watched is broken. Not sure anything happens ever or if something weird happens that I can't identify.

### Additional information

Link to support request: https://forums.trakt.tv/t/trakt-support-request/116733

rectifyer:
# Summary

The "Display Watched" (and "Display Watchlisted") toggles did nothing on a person's credits pages. This honors both filters client-side for movie and show credits.

Closes #2943.

# Root cause

The person credits pages send `ignore_watched` / `ignore_watchlisted` to `/people/:id/movies` and `/people/:id/shows`, but those endpoints (worker-served, `getPersonMovies` / `getPersonShows`) apply only the generic `filters()` middleware and never filter by watched or watchlisted status — the SDK contract doesn't even define the ignore params for `people.*`. So the toggles had no effect. Same class of bug as the calendar (#2941), different surface.

# Fix

A pure predicate `isCreditHidden` decides whether a credit should be hidden given the two toggles and the user's `history` / `watchlist`. `useCreditsList` folds `filter$`, `history`, and `watchlist` into the credits pipeline and drops hidden credits before the intl overlay, so it works on the movies page, the shows page, and combined `media` mode. When a toggle is off, or its backing store is still loading, nothing is filtered.

- Movies: watched via `history.movies`, watchlisted via `watchlist.movies`.
- Shows: watched via `history.shows`, watchlisted via `watchlist.shows`.

# Show "watched" semantics

A credit only carries a base `MediaEntry`, which has no total episode count, so the fully-watched check used elsewhere (`useIsWatched`) can't be computed here. A show credit is treated as watched when the user has any watch history for it (`history.shows.has(id)`) — the same way discover's `ignore_watched` treats shows. If we want strict fully-watched here, the credits mapper would need to carry the aired-episode count; that can be a follow-up.

# Tests

`isCreditHidden.spec.ts` covers movie/show against watched/watchlisted, both toggles off, and pass-through while history/watchlist are still loading.

# Notes

If every credit in a crew position is hidden, that position still appears (empty) in the position selector. Easy to filter out as a follow-up if desired.

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -89,6 +89,7 @@ module.exports = {
         'notes',
         'out-now',
         'parameter',
+        'person',
         'player',
         'popover',
         'popular',
--- a/projects/client/src/lib/sections/lists/stores/_internal/isCreditHidden.spec.ts
+++ b/projects/client/src/lib/sections/lists/stores/_internal/isCreditHidden.spec.ts
@@ -1,0 +1,123 @@
+import type { UserWatchlist } from '$lib/features/auth/queries/currentUserWatchlistQuery.ts';
+import type { UserHistory } from '$lib/features/auth/stores/useCurrentUserHistory.ts';
+import type { MediaEntry } from '$lib/requests/models/MediaEntry.ts';
+import type { MediaType } from '$lib/requests/models/MediaType.ts';
+import { describe, expect, it } from 'vitest';
+import { isCreditHidden } from './isCreditHidden.ts';
+
+function media(type: MediaType, id: number): MediaEntry {
+  return { id, type } as unknown as MediaEntry;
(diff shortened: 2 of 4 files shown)


seferturan:
Similar to this one: https://github.com/trakt/trakt-web/pull/2942#issuecomment-5024192669
We should fix the filtering server side, not filter these client side.

PR #2945, opened 2026-07-17
Episode: restore where to watch section in the episode drawer

### LINKED ISSUE #2897: Episode Drawer: Missing Content
### Code of Conduct

### Where did you notice the design improvement?

v3 Desktop, episode drawer.

### What do you see as an improvement?

Because somehow it was decided to stuff everything into drawers, we now no longer have full episode pages but episode drawers. So on page load, both the series and the episode content has to load, but the episode content you actually came for is now crammed into a fraction of the available screen real-estate.
Other than is just looking overstuffed and unclear and the whole "drawer within a drawer" mess for stuff like social activity, the most problematic consequence I found is missing content. Specifically, Where to Watch for the episode is gone, as are the remaining episode details (writers, plays...).

### How should it ideally feel or look instead?

At the very least, bring back the missing content.

### Screenshots or visual references

_No response_

### Additional context

_No response_

rectifyer:
# Summary

Restores the **Where to Watch** section in the v3 episode drawer. When episode pages became drawers, the streaming list was dropped; this brings it back, positioned like the mobile show/movie layout — directly under the episode overview, above the tabs. Closes #2897.

# Changes

- Add a `useEpisodeStreamOn` hook to load per-episode streaming data (mirrors the existing episode-page loader).
- Render `WhereToWatchList` in `EpisodeDrawerHost`, gated on the loaded episode, using the `inline` list variant so the empty state fits the narrow drawer instead of overflowing on desktop.
- Open "view all" as a locally-mounted, episode-scoped `WhereToWatchDrawerHost` (elevated/stacked) instead of the `view=` URL param — matching how the drawer already handles ratings/social/history, so it stays episode-scoped and doesn't replace the drawer.
- Thread an optional `onDrilldown` callback through `SectionList` → `ListHeader` → `ListTitle` (additive; the drilldown renders a button that fires the callback). Reuses the `onDrilldown` pattern `RatingList` already uses in this drawer.
- Add an `elevated` prop to `WhereToWatchDrawerHost` so it can stack on top of the episode drawer.
- Align the section's title and cards with the drawer's tab content (the People title) and tighten the gap to the overview.

# Notes

Writers and plays (also called out in the issue) are already present — plays in the Info tab stat row, writers under the People section's crew toggle — so this focuses on the genuinely missing Where to Watch section.


addisonxx:
Another missing section is the external links that used to exist in the details popup for each episode.

seferturan (approved):
Applied some fixups:

- Moved it to the info tab.
- Added a skeleton for it so content doesnt jump around.
  - Gave it it's own isAired check when going from aired to un aired episodes to not show the skeleton.
  - Caveat: this works best when data is cached; so cold hits might still show a skeleton for unaired episodes the first time. We can fix it more properly later on.

PR #2978, opened 2026-07-21
List: type a rank to reorder list items

rectifyer:
# Overview

Adds fast rank editing to the list-reorder drawer, on top of the existing drag-and-drop. Applies to both reorder drawers — the all-lists drawer and the per-list/watchlist drawer — since both render the shared `ReorderDrawer`. No API or model changes; everything persists through the existing reorder request.

# Desktop

The rank number on each row is now an inline editable field. Clicking (or tabbing into) it selects the value, so you can type a new position and press Enter or blur the field to move the item there. Escape reverts. The value clamps to `1…length` and every other row renumbers automatically. Next to the drag handle there are move-to-top and move-to-bottom icon buttons (iOS-style `arrow.up/down.to.line` glyphs), disabled when the row is already first or last.

https://github.com/user-attachments/assets/123b5a09-0cc2-42de-8587-35d4839f9378

# Mobile

Tapping the rank opens a bottom-sheet drawer with a number input and Cancel/Apply, instead of an anchored popover that the on-screen keyboard would cover. The same move-to-top and move-to-bottom buttons are available.

https://github.com/user-attachments/assets/95999c5d-97e0-4a89-9a28-6b022cab9133

# Shared behaviour

Drag-and-drop is unchanged and still live-updates the rank numbers as you drag, including the floating drag ghost. Every way of moving an item — the inline field, the mobile sheet, the top/bottom buttons, and drag-and-drop — routes through one code path that scrolls the moved row into view and briefly highlights it with a ~1.5s fade, so it is easy to see where the item landed. Accessible names are provided on the rank control, the animation respects `prefers-reduced-motion`, and the layout stays RTL-safe.

# Implementation notes

- New `RankEditor` component splits desktop (inline field) vs mobile (bottom sheet) via `useMedia`; the nested sheet uses `Drawer` with `elevated` so it stacks correctly above the reorder drawer.
- New `ArrowUpToLineIcon` / `ArrowDownToLineIcon` icons.
- A single `moveItemToRank` helper backs all move actions; `highlightMovedRow` handles the scroll + flash.
- Reuses the previously-staged i18n keys (move to top/bottom, position prompt); no new message keys were added.

### DIFF
--- a/projects/client/src/lib/sections/lists/user/_internal/ReorderDrawer.svelte
+++ b/projects/client/src/lib/sections/lists/user/_internal/ReorderDrawer.svelte
@@ -8,14 +8,20 @@
   import * as m from "$lib/features/i18n/messages.ts";
   import CrossOriginImage from "$lib/features/image/components/CrossOriginImage.svelte";
   import { MEDIA_POSTER_PLACEHOLDER } from "$lib/utils/assets.ts";
+  import { clamp } from "$lib/utils/number/clamp.ts";
+  import { time } from "$lib/utils/timing/time.ts";
+  import { onDestroy, tick } from "svelte";
   import { flip } from "svelte/animate";
   import { cubicOut } from "svelte/easing";
   import type { ReorderableListItem } from "./models/ReorderableListItem.ts";
+  import { moveReorderableItem } from "./moveReorderableItem.ts";
+  import RankEditor from "./RankEditor.svelte";
   import { type DragGhost, reorderDrag } from "./reorderDrag.ts";
   import {
     itemOrderSignature,
     sortReorderableItems,
   } from "./reorderListItems.ts";
+  import { scrollIntoViewWhen } from "./scrollIntoViewWhen.ts";
 
   const {
     items,
@@ -37,6 +43,8 @@
 
   const { confirm } = useConfirm();
 
+  const flashDuration = time.seconds(1.5);
+
   let localOrder = $state<{
     signature: string;
     items: ReorderableListItem[];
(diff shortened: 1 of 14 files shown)


projects/client/src/lib/sections/lists/user/_internal/ReorderDrawer.svelte
```diff
@@ -160,6 +160,65 @@
     dragGhost = null;
   }
 
+  function currentRankOf(key: string) {
+    return orderedItems.findIndex((item) => item.key === key) + 1;
+  }
+
+  function handleRankKeydown(
```
seferturan: The key handling here I think is strange for mobile (i.e. I think on mobile there's usually an explicit 'commit' button for such things), and the `event.currentTarget.select()` below seems to pop up the text selection popup menu (at least in responsive mode on chrome) 😅
Maybe for v0 we can mimic the iOS approach for now?

seferturan (changes requested):

projects/client/src/lib/sections/lists/user/_internal/ReorderDrawer.svelte
rectifyer: @seferturan Code updated to handle mobile view more like iOS with a slide up drawer to edit the rank. I also added a quick move first/last icon since iOS has that too. Lastly, added a slight flash animation to visually confirm where the item was moved to.

seferturan: Love that flash effect :D

seferturan (approved):
Added some fixups to make it a bit more similar to the iOS version:

- The while thing is now under a popup menu.
- The ranks are tapable/clikable buttons.
- 3 options in there: `move to top`, `move to...`, `move to bottom`

Reasoning:
- It was visually quite heavy having so many inputs visible at the same time.
- The move to top/bottom buttons on the item itself could end up being confusing, since they don't have labels. E.g. it could be easy to mistake them for moving them one position up or down.

Maybe we can do a follow up here to improve the desktop experience. @ElMagnea, once deployed, when you can, please take a look.

PR #2983, opened 2026-07-22
fix(search): NaN episode counts in search results

### LINKED ISSUE #2975: Wrong Show Indicator in Search Results
### Code of Conduct

v3 Desktop, Search Result Page

Completed Shows have the "In Progress" indicator on their posters:
Image

On detail pages, the indicator is correct:
Image

### Additional information

_No response_

### LINKED ISSUE #2976: NaN Episode Count in Search Results
### Code of Conduct

v3 Desktop, Search Result Page, "Shows" Mode

All episode counts render as NaN:

Image

### Additional information

_No response_

rudf0rd:

Shows in search results render "NaN eps." because `toShow` maps `aired_episode_count`, a field live Typesense Show documents never populate. `mapToShowEntry` then turns the missing value into a literal `NaN`. Now falls back to `episode_count`, which exists on every Show document (it's the same field `getMedia` uses to detect show hits).

On top of the count fallback, `dedupe` now keeps the **first** occurrence per key instead of the last. Search merges exact → trending → fuzzy; last-wins meant a fuzzy Typesense hit silently replaced the richer API entry for the same show — displayed at the exact match's rank but carrying Typesense data. API entries carry the true `aired_episodes`, so they now win on data as well as position. This is what made the watched indicator flip between the media and shows toggles: the shows-only fuzzy pass always contained the show, while the movie+show union often cut it from the top 50, so each mode surfaced a different source's copy.

Fixes #2976. Fixes #2975.

**Caveat — this is ~80%, not 100%:** a currently-airing show that surfaces *only* in the fuzzy pass (no exact-title match, not trending) still falls back to `episode_count`, the total *ordered* count — so a caught-up viewer can read as In Progress. Popular and airing shows almost always ride the exact or trending path, so most real searches get the true aired count. Closing the last gap needs `aired_episode_count` populated in the Typesense index and exposed via the `search:media` preset (server-side follow-up).

Before (from #2976):

NaN episode counts

After:
CleanShot 2026-07-22 at 10 52 36@2x

- Verified against the live cluster: every Show hit for a sample fuzzy query lacks `aired_episode_count`, all carry `episode_count`.
- Manual repro with local Typesense creds: fuzzy query in Shows mode shows NaN before, real counts after.
- Verified the #2975 repro: searching "lucky" and toggling media ↔ shows now shows the same correct watched check on Lucky (2026) in both modes — previously shows mode read In Progress because the fuzzy hit (total count 7) clobbered the exact entry (aired count 3).
- `dedupe.spec.ts` updated to pin first-occurrence semantics; the other two `dedupe` call sites each dedupe a single array, so the change is inert there.
- Note: `searchMediaQuery.spec.ts` is currently skipped, so no spec covers this mapper.


projects/client/src/lib/utils/array/dedupe.ts
```diff
@@ -7,9 +7,14 @@ export function dedupe<T>(
 ): T[] {
   const map = new Map<KeygenValue, T>();
 
+  // First occurrence wins: earlier source arrays take precedence over later
+  // ones, so callers can order sources from most to least authoritative.
```
vladjerca: @rudf0rd all of these comments are unnecessary. Remove them.

vladjerca (changes requested):

seferturan (approved):
Applied some fixups:

- Schemas now mirror the new Typesense docs. Ours still described the old documents, so
  `toShow`/`toCommonMedia`/`toIds`/`toPerson` read ~30 fields that no longer exist. All resolved to
  `undefined` behind a `?? NaN` / `?? 'TBD'` / `?? 0`, NaN eps and missing runtimes were just the visible one.
- Dropped the `episode_count` fallback. `aired_episode_count` and `runtime` are indexed again. Verified against live, no more NaNs or missing
  runtimes 👌
- Hits without a slug are dropped. Slug is optional on every collection and `MediaEntry`/`PersonSummary`
  both require one, so those used to fail validation and error the whole query.
- Trending only fires for the active mode now. `?m=movie&q=dexter` was pulling the show in from
  `/recent_by_id/global/shows`. One less request per search too.

Issue #1710, opened 2026-02-17
Display list like counts


seferturan:
We can now like/unlike lists, but don't show the amount of likes a list has. We should incorporate that into the button.

PR #2984, opened 2026-07-22
fix(list): stop media detail lists rearranging on like/unlike

### LINKED ISSUE #2973: Un/Liking List on Media Detail Page Rearranges Element
### Code of Conduct

v3 Desktop, some media detail pages (not everywhere), e.g. https://app.trakt.tv/movies/hell-or-high-water-2016?mode=media

After liking and unliking some (doesn't work with all) lists, the wrapper element seems to unmount completely and re-renders with the unliked list in a different spot:

https://github.com/user-attachments/assets/6d7682a5-9047-4801-95e3-6901574744ff

### Additional information

_No response_

rudf0rd:

Fixes #2973. Liking or unliking a list on a media detail page made the whole lists section remount and shuffle the list into a new spot.

The cause: `movieListsQuery` / `showListsQuery` declared `InvalidateAction.List.Like`, so every like/unlike refetched them — and since they fetch with `sort: 'popular'` and liking changes popularity, the refetch reorders the section. The invalidation isn't needed for the heart state; that comes from `currentUserLikesQuery`, which keeps its own `List.Like` invalidation. So this just drops the token from both queries.

**Trade-off:** the invalidation was originally added (d0579a8ef, #1710) so the like *count* on these cards refreshes after you like. With it removed, the count you see on media detail list cards is stale by one until the 30-min TTL — heart state still updates instantly, all other surfaces (liked lists, personal lists, list detail) keep their invalidations. IMO an off-by-one count beats the section jumping around; if we want live counts back we can optimistically patch `likeCount` in `useLikeList` as a follow-up.

Repro video on #2973 shows the before. 

After:

https://github.com/user-attachments/assets/ade9547c-6671-42ea-ba3f-bb837a1e7eee

- `movieListsQuery.spec.ts` and `showListsQuery.spec.ts` pass.
- Manual: on `/movies/hell-or-high-water-2016?mode=media`, like/unlike lists in the lists section — no remount, no reorder, no refetch of `/movies/.../lists` in the network tab.

### DIFF
--- a/projects/client/src/lib/sections/summary/components/lists/useListSummary.ts
+++ b/projects/client/src/lib/sections/summary/components/lists/useListSummary.ts
@@ -1,6 +1,8 @@
+import type { MediaListSummary } from '$lib/requests/models/MediaListSummary.ts';
 import type { MediaType } from '$lib/requests/models/MediaType.ts';
 import { movieListsQuery } from '$lib/requests/queries/movies/movieListsQuery.ts';
 import { usePaginatedListQuery } from '$lib/sections/lists/stores/usePaginatedListQuery.ts';
+import { useStableArray } from '$lib/sections/lists/stores/useStableArray.ts';
 import { combineLatest, map } from 'rxjs';
 import { showListsQuery } from '../../../../requests/queries/shows/showListsQuery.ts';
 import { MAX_LISTS } from './_internal/constants.ts';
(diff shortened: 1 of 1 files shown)


seferturan (approved):
Applied a fixup; instead of removing the invalidations, it now uses stable array.

PR #2987, opened 2026-07-22
fix(list): seed sort params from URL to avoid duplicate first fetch

rudf0rd:

The sorting stores (`useListSorting`, `useUpNextSorting`, `useUserListsSorting`) boot with `null` sort params; the URL's `sort_by`/`sort_how` only sync in via a post-mount `$effect` in `ListSortActions`. Result: refreshing any sorted list page fetches page 1 twice — once with the fallback sort (briefly rendering the wrong order), then again with the URL's sort.

Fix: seed each store from `page.url.searchParams` at construction. The effect then pushes identical values, the query key never changes, and only one request fires.

Refreshing `/users/me/watchlist?sort_by=rank&sort_how=asc`, before:

```
GET .../watchlist/movie,show?...&sort_by=added&sort_how=desc   ← wasted
GET .../watchlist/movie,show?...&sort_by=rank&sort_how=asc
```

After: only the second call.

- `useListSorting.spec.ts`: new spec asserting the store's first emission already carries the URL's sort.
- Manual: refresh a list page with sort params in the URL — one request in the network tab, no re-order flicker.

### DIFF
--- a/projects/client/src/lib/sections/lists/progress/useUpNextSorting.spec.ts
+++ b/projects/client/src/lib/sections/lists/progress/useUpNextSorting.spec.ts
@@ -2,12 +2,16 @@ import { FeatureFlag } from '$lib/features/feature-flag/models/FeatureFlag.ts';
 import { useFeatureFlag } from '$lib/features/feature-flag/useFeatureFlag.ts';
 import { renderStore } from '$test/beds/store/renderStore.ts';
 import { filter, firstValueFrom, map, take } from 'rxjs';
-import { beforeEach, describe, expect, it } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
 import { useUpNextSorting } from './useUpNextSorting.ts';
 
 describe('useUpNextSorting', () => {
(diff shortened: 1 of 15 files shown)


seferturan (approved):
I applied some fixups:
- Added `useSortParams`, deriving `sort_by`/`sort_how` from `useParameters().url` instead of every store keeping its own `BehaviorSubject`.
- Dropped `update` from `useListSorting`, `useUpNextSorting` and `useUserListsSorting`, plus the `onUpdate` prop and `$effect` in `ListSortActions`, and unwired `update` from the 8 call sites. URL is read in one place now instead of two.
  - `update` existed because these stores are plain `.ts` and can't reactively read `page.url`, so the read lived in a component and got pushed in. `useParameters().url` gives us that reactive URL as an observable now, so the bridge is redundant.
- Side effect: the pages wrapping these stores in `$derived` were dropping `?sort_by=` entirely (store re-runs, fresh subject, loses whatever the effect pushed). Those honour the URL now too.
- Updated tests accordingly.

PR #2992, opened 2026-07-23
feat(drawer): grouped-card reskin for drawers and popup menus

ElMagnea:
## Visuals
Screenshot 2026-07-23 at 11 50 40
Screenshot 2026-07-23 at 11 50 52
Screenshot 2026-07-23 at 11 55 52
Screenshot 2026-07-23 at 11 56 15
Screenshot 2026-07-23 at 11 57 04

## What & why

The drawers and popup menus were visually dated — filled "pill" rows with gaps, no shared surface, inconsistent chrome. This reskins them into a single **grouped card** language (elevated surface, flush hairline-divided rows, subtle hover/press, eyebrow titles) applied consistently across the app.

## The reusable primitive

Every menu now routes through one of two components instead of hand-rolling its own list:

- **`DropdownGroup`** — grouped card that flattens child `DropdownItem`s (filled pills → flush divided rows). Containers pass items as opaque snippets, so the flatten is done via scoped overrides in one place.
- **`OptionList` / `OptionListItem`** — the same card as first-class components (icon · title · optional subtitle · trailing slot), used where items aren't `DropdownItem`s.

## Coverage

| Surface | Treatment |
| --- | --- |
| Card "···" / recommendation / action menus (mobile drawer + desktop portal) — `PopupMenu` | grouped card |
| Profile overflow, streaming-service, subscription menus | grouped card |
| Summary "···" menu (mobile drawer + desktop slider) | grouped card |
| Manage Lists | grouped card (bookmark stays trailing) |
| Track… (mark-as-watched) | grouped card |
| Watch-until-here mode picker | grouped card, active row shows a ✓ |
| Sort drawer | grouped card via `OptionList` + two-line subtitles |

Audited every component with 2+ `DropdownItem`s — no ungrouped menu containers remain.

## Also in here

- Shared `Drawer`: eyebrow title as the default (`titleStyle="prominent"` opt-out for content-name titles like show/episode names, auto-forced when a drilldown title is set); refined drag handle; more handle-to-title spacing.
- Desktop dropdowns no longer scroll/clip — always show every option.
- Opaque card surface so floating menus don't turn transparent over artwork.
- Light-theme contrast pass: distinct near-white surface + defining border (new `--color-option-list-border`) so cards don't read as white/transparent.
- New tokens in both themes: `--color-drawer-drag-handle`, `--color-option-list-background` / `-border` / `-separator` / `-highlight`.
- 4 new i18n subtitle keys for the up-next sort options.

## Verification

`deno fmt` clean · `svelte-check` 0 errors / 0 warnings · `vitest` 2414 passed / 7 skipped.

## Screenshots

_To be added._


vladjerca:
@ElMagnea can we also reorder actions on recommendations for example?

Not have to "why this?" Next to "ignore"

Please do this on a separate PR 🙏

seferturan (approved):
Applied some fixups:

- Reduced the scope, this now mostly focuses on the menu item groups in drawers. Popup menu changes to be tackled separately.
  - Kept the full height dropdown though, so no scrolling inside the popup menu.
  - Mobile popup menus still open in a drawer with the new grouped card.
  - Dropped the eyebrow title treatment. Drawer titles are back to normal case everywhere.
  - Drag handle: only kept the wider bar. Reverted the dimmer colour and the smaller tap area, those made it harder to grab.
- Fixed click zones in the grouped rows. Padding lived on the li while the <a> was the actual target, so the padded area did nothing. Padding now sits on the link.
  - Also fixes the inline dead zone that was already there for linked dropdown rows.
- Moved the grouped row styling into DropdownItem itself. DropdownGroup is just the card shell now, it no longer sets styling on its children.
- Selected row was barely visible in dark mode. New --color-option-list-selected token, mixes towards the text colour so it lifts on dark and tints on light.
- Sort option subtitles now follow the sort direction. Newest first / Oldest first, Most episodes left / Fewest episodes left.
  - Least episodes left was just wrong before, desc sorts most first.
  - Added specs for the direction aware subtitles.
- SummaryActionsSlider used a :global() override on the group. Added a style="flat" variant instead.
- Removed DropdownList. It had no consumers left, only the design system page. That page now shows DropdownGroup + DropdownItem.

PR #2994, opened 2026-07-23
fix(search): filter trending results by active mode

### LINKED ISSUE #2977: Mode Toggles Don't Work in Search Results
### Code of Conduct

v3 Desktop, Search Result Page

The mode toggles do apply some sort of filter, but some movies get listed as shows and vice versa.

Searching for Avatar, the Netflix show is listed as a movie:
Image

While it also shows up as a show, some movies are also mixed in:
Image

### Additional information

NaN episode counts from screenshot via #2976

rudf0rd:
## Scene Setting

The mode toggles on search results filter the typed search queries, but the trending search overlay always fetches movies **and** shows and merges them in unfiltered — so shows leak into Movies mode and vice versa. This filters trending entries by the active mode before the merge. The trending cache stays shared across modes, so toggling never triggers a refetch.

Fixes #2977

## Show, Don't Tell

**Before** (from the issue — the Avatar show listed under Movies, movies mixed into Shows):

show listed as movie
movies mixed into shows

**After**

CleanShot 2026-07-23 at 10 43 26@2x
CleanShot 2026-07-23 at 10 44 06@2x

## Testing

- New regression specs in `useSearch.spec.ts`: media mode includes trending entries of every type; movie mode excludes other types. Verified the movie-mode spec fails without the fix.
- New `searchTrendingQuery.spec.ts` plus MSW handlers/fixtures for the trending endpoints (previously unmocked; the catch-all search mock also matched case-sensitively and could never hit, since `useSearch` lowercases terms).

### DIFF
--- a/projects/client/src/lib/features/search/useSearch.spec.ts
+++ b/projects/client/src/lib/features/search/useSearch.spec.ts
@@ -1,7 +1,9 @@
+import { MovieHereticResponseMock } from '$mocks/data/summary/movies/heretic/response/MovieHereticResponseMock.ts';
+import { ShowSiloResponseMock } from '$mocks/data/summary/shows/silo/response/ShowSiloResponseMock.ts';
 import { renderStore } from '$test/beds/store/renderStore.ts';
 import { waitForEmission } from '$test/readable/waitForEmission.ts';
 import type { Subscription } from 'rxjs';
-import { afterEach, describe, expect, it, vi } from 'vitest';
+import { afterEach, beforeAll, describe, expect, it, vi } from 'vitest';
 import { useSearch } from './useSearch.ts';
 
 const { trackSpy } = vi.hoisted(() => ({ trackSpy: vi.fn() }));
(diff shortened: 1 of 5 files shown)


projects/client/src/lib/features/search/useSearch.ts
```diff
@@ -130,7 +130,10 @@ export function useSearch() {
               items: dedupe(
                 (item) => item.key,
                 (exactResults as MediaSearchResult).items,
-                trendingResults?.items ?? [],
+                (trendingResults?.items ?? []).filter(
+                  (item) =>
+                    currentMode === 'media' || item.type === currentMode,
+                ),
```
vladjerca: @rudf0rd this is not the correct fix, we have separate requests by media type filtering is just hiding the root cause.

vladjerca (changes requested):

projects/client/src/lib/features/search/useSearch.ts
rudf0rd: @vladjerca updated to use the endpoints. let me know if the Promise.all is not the right direction.  if it's not, i can further break it down

seferturan (approved):

PR #2995, opened 2026-07-23
feat(toast): action-confirmation toasts with undo & change-list

ElMagnea:
## Visuals

Marking something as watched by mistake
Screenshot 2026-07-23 at 15 09 29

Accidentally removing an item from a watchlist
Screenshot 2026-07-23 at 21 34 12

When adding an item to your watchlist, you can pick **another** list right after _(if need be)_
Screenshot 2026-07-23 at 21 34 18

## What

Adds a global **action-confirmation toast** engine, gated behind a new `ActionConfirmations` preview flag (**off by default**; on for directors). The shared media mutation hooks fire a confirmation toast after each action, with an inline undo / "change list" affordance.

| Action | Toast | Affordance |
|---|---|---|
| Add to watchlist | "*Title* was added to your watchlist" | **Change list** → opens the manage-lists drawer |
| Remove from watchlist / favorites / history | "*Title* was removed…" | **Undo** |
| Rate / remove rating | "You rated *Title* X/10" / "Rating removed" | **Undo** |

## How

- New feature module `lib/features/action-toast/` — a module-store singleton (mirrors `useDismissals` / `markAsWatchedDrawerStore`) + `ActionToastHost` mounted once in the layout, reusing the existing `Snackbar` primitive.
- **Undo runs the sibling hook method** (via `executeOrEnqueue`), not the raw `*Request` — so it stays offline-aware, cache-invalidating, and consistent with the optimistic read overlay.
- Global manage-lists drawer opener (`manageListsDrawerStore` + provider) so "change list" can open the drawer from anywhere.
- Snackbar polish (shared): bold inline title (`MessageWithBold`), a **theme-aware** outline "pill" action (stroke token added to both `modes.scss` mixins), a hairline border, and a bottom fly-in/out animation (`|global`; respects `prefers-reduced-motion`).

## Notes / scope

- Behind a flag → safe to merge; no behavior change until toggled on.
- Verified in **both light and dark** themes.
- **Deferred (follow-ups):** `useRemoveFromHistory` (row-level, lossy undo) not wired; ratings "Change" (needs a global rating drawer); the outline-button look is scoped to the toast — a global button restyle is its own pass.

## Verification

`deno fmt` ✓ · `svelte-check` 0/0 ✓ · `vitest` 2418 passed ✓ · `i18n:check` ✓. Undo is covered by tests that assert the **real reverse request** fires (watchlist / favorites / history).

### DIFF
--- a/projects/client/i18n/meta/en.json
+++ b/projects/client/i18n/meta/en.json
@@ -10645,6 +10645,92 @@
     "label_genres_drawer_all": {
       "default": "All Genres",
       "description": "Section label above the full genre catalog in the genres drawer."
+    },
+    "action_toast_added_to_watchlist": {
+      "default": "<b>{title}</b> was added to your watchlist",
+      "description": "Confirmation toast shown after adding a title to the watchlist. {title} is the movie or show title.",
+      "variables": {
+        "title": {
+          "type": "string"
+        }
+      }
+    },
+    "action_toast_added_to_watchlist_generic": {
+      "default": "Added to your watchlist",
+      "description": "Confirmation toast shown after adding to the watchlist when no single title is available (e.g. bulk actions)."
+    },
+    "action_toast_removed_from_watchlist": {
+      "default": "<b>{title}</b> was removed from your watchlist",
+      "description": "Confirmation toast shown after removing a title from the watchlist. {title} is the movie or show title.",
+      "variables": {
+        "title": {
+          "type": "string"
+        }
+      }
+    },
+    "action_toast_removed_from_watchlist_generic": {
+      "default": "Removed from your watchlist",
+      "description": "Confirmation toast shown after removing from the watchlist when no single title is available."
+    },
+    "action_toast_action_change_list": {
+      "default": "Change list",
+      "description": "Inline link in the watchlist confirmation toast that opens the manage-lists drawer to move the item to a different list."
+    },
+    "action_toast_label_change_list": {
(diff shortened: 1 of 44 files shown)


michaldrabik:
- for actions that already have a confirmation dialog "Are you sure..." adding this seems a bit excessive and just noise?
- for Rating Prompt UI can we do this somehow so it does not make everything taller and break symmetry? :P Maybe just an undo icon button next to X is enough, should be clear.
- also just thinking now that Rating Prompt is not something strictly related to your active action of marking as watched so that might not be the best place to undo anything (might have scrobbled, checkin might have ended, other client might have marked etc.). 🤔

seferturan:
Yeah, I'm removing it from the rating prompt, since that's not really a snackbar that's shown as the result of a direct action.
I also need to look a bit more into the undo, i.e. it needs to undo only the specific action, and not clear the entire history for an item 😅

seferturan (approved):
Some changes:

- This is never added to the rating prompt. Since that prompt is not always a result of a direct action, and can be shown way later.
- I don't like the snackbar appearing in so many cases, especially when the action that would trigger it is still there on screen (i.e. it would be very easy to just undo it from there where your finger/cursor is already at). So for now I changed it to only ever be triggered from actions in popup/dropdown menus.
- Generalized the list drawer code, all list drawers now work in the same way.
- Removed the code that was no longer relevant due to those changes.

PR #3006, opened 2026-07-26
Better translation of "link_text_preview_settings" on nl-NL.json

thestefbauwens:
Better translation of "link_text_preview_settings", because 'Voorbeeld' seems like 'Example'. Something like 'Bèta' could be an option too.


michaldrabik:
Hi @thestefbauwens for everything related to translations please join our CrowdIn at: https://translate.trakt.tv/
You can then open issues or suggest strings for your language specifically. Thanks!

PR #3008, opened 2026-07-27
Reorg Account Settings

Marius-TV:
Move the Account Details sections out of the settings Profile card into a new AccountSettings component, routed at /settings/account and listed below General in the settings navigation for both desktop and mobile.


seferturan (approved):
Applied some fixups:

- Removed the dead `label` code; extracted some duplicate coce.
- Pruned stale translations.

PR #3012, opened 2026-07-27
List: ignore leading articles when grouping title sort

### LINKED ISSUE #3009: Use of Article in Title
### Code of Conduct

In V2 the use of "article" in the title of a show or movie was ignored so that titles sorted in logical order not in true alphabetical order.

What we currently have in V3 is this sort order where the article is **included**:
**For example:**
Alex: The Life of a Child
Alice, Sweet Alice
All About Bette
All Over the Guy
The Alchemist
The Alice

In V3 can we please go back to **ignoring** the article (in this example it is "The") and have the sort, search etc. sort in a logical order not in true alphabetical order.
**For example:**
The Alchemist
Alex: The Life of a Child
The Alice
Alice, Sweet Alice
All About Bette
All Over the Guy

I don't think there is an alternative to what is in V3 and what was in V2

### Additional context

_No response_

rectifyer:
# What

When a list is sorted by title, titles that begin with a leading article ("A", "An", "The") now file under their first significant letter — "The Beatles Anthology" groups under **B**, not **T** — so the alphabetical section dividers line up with the sort order.

Fixes #3009.

# Why

The API already returns list items ordered with leading articles ignored, but the web client still computed each alphabetical divider from the raw first character of the title. The result was a single broken "T" bucket holding every "The …" show, completely out of step with the order the items actually came back in (see the screenshot on the issue).

# How

Added a `toSortableTitle` utility that strips a leading English article ("a", "an", "the") — case-insensitive, and only when whitespace follows it — mirroring the normalization the API applies when it sorts, so "Andromeda" and a title that is literally "The" are left untouched. The `title` branch of `formatSortValue` now derives the divider letter from the article-stripped title, which fixes every list surface that groups by first letter (user lists, watchlist, favorites, progress) from a single place.

# Testing

- Added unit coverage for `toSortableTitle`: the three articles, case-insensitivity, "only the first article", and the untouched edge cases ("Andromeda", "Theory", literal "The").
- Updated the `formatSortValue` title specs to expect article-stripped grouping ("The Wire" → W, "The Boys" → B).

### DIFF
--- a/projects/client/src/lib/components/lists/grid-list/GridList.spec.ts
+++ b/projects/client/src/lib/components/lists/grid-list/GridList.spec.ts
@@ -1,0 +1,63 @@
+import { render } from '@testing-library/svelte';
+import { createRawSnippet } from 'svelte';
+import { describe, expect, it } from 'vitest';
+import GridList from './GridList.svelte';
+
+type TestItem = { key: string; title: string };
+
+function renderGridList(items: TestItem[]) {
+  return render(GridList<TestItem>, {
+    props: {
+      id: 'test-grid-list',
+      title: undefined,
+      items,
+      groupBy: (item: TestItem) => item.title[0] ?? '#',
+      item: createRawSnippet((item: () => TestItem) => ({
+        render: () => `<div class="test-item">${item().title}</div>`,
+      })),
+    },
+  });
+}
+
+function toRenderedSequence(container: HTMLElement) {
+  const nodes = container.querySelectorAll(
+    '.trakt-letter-group-header, .test-item',
+  );
+
+  return Array.from(nodes, (node) => node.textContent?.trim());
+}
+
+describe('GridList', () => {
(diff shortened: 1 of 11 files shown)


seferturan (dismissed):
I applied some fixups:

  - Grouping was reading the localized title, so headers were wrong in non-English locales, and it stripped the leading "A" off things like "A ciegas". Now keeps the untranslated entry around through the intl overlay and groups off that.
  - `GridList` bucketed groups into a `Map` keyed by letter, so every later "The ..." got pulled into the first slot that letter appeared in. That's why `T` sits right under `1`, and why the real T titles never show up as you scroll (they get appended way above the fold). Groups consecutive runs now.
  - Verified against the lists from #3009 with playwright, rendered headers match the API order exactly now.
  - Left a fixme: we shouldn't be re-deriving the sort key client side, the server should just hand us a `sort_title`.

seferturan (approved):

PR #3013, opened 2026-07-27
Spoiler: respect spoiler setting in episode button titles

### LINKED ISSUE #2972: Button Titles Don't Respect Spoiler Setting
### Code of Conduct

v3 Desktop with mouse

Spoiler-hidden episode titles are visible as track button title in Continue Watching & episode drawer:

Image

### Additional information

_No response_

rectifyer:
# Summary

Closes #2972. Spoiler-hidden episode titles were still visible in the track button's tooltip / `aria-label` in Continue Watching and the episode drawer. Cards blur the title visually via `<Spoiler>`, but the raw `episode.title` was folded into plain button strings (tooltips, `aria-label`s, popup headers) that aren't blurred, so the spoiler leaked there.

# Fix

Added `useSpoilerFreeEpisodeTitle`, a small hook mirroring the existing `useEpisodeSpoilerImage` pattern. It reuses `useMediaSpoiler` (unwatched + spoiler preference → hidden, the same signal that drives the visual blur) and swaps in a spoiler-free season/episode label (e.g. `S1 • E5`) when hidden, otherwise passing the real title through unchanged.

Applied it everywhere the title was folded into a button string on a surface that blurs the title:

- Continue Watching track button (the one in the issue screenshot) and all cover-card "next" buttons — `EpisodeItem.svelte`
- Episode drawer track button — `EpisodeInfoHeader.svelte` (the drawer only shows the S/E number, never the title heading)
- Continue Watching popup "Mark as watched" item — `UpNextItem.svelte`
- Continue Watching card "⋯" popup `aria-label` + mobile drawer header — `EpisodeCard.svelte`
- Summary-style card "⋯" popup `aria-label` — `MediaSummaryCard.svelte`

# Scope

The full episode summary page is intentionally left untouched: it renders the title openly as a heading, so masking only its button tooltip would be inconsistent. When an episode is watched, or spoilers are off, behavior is unchanged and the real title flows through as before.

# Testing

Added a colocated spec `useSpoilerFreeEpisodeTitle.spec.ts` covering both branches (title shown when not hidden, season/episode label when hidden). Ran the spoilers feature suite via `vitest` directly — new spec 2 passed, whole spoilers feature 15 passed.

### DIFF
--- a/projects/client/src/lib/features/spoilers/useSpoilerFreeEpisodeTitle.spec.ts
+++ b/projects/client/src/lib/features/spoilers/useSpoilerFreeEpisodeTitle.spec.ts
@@ -1,0 +1,100 @@
+import { episodeSubtitle } from '$lib/utils/intl/episodeSubtitle.ts';
+import { clone } from '$lib/utils/object/clone.ts';
+import { deepAssign } from '$lib/utils/object/deepAssign.ts';
+import { EpisodeSiloMappedMock } from '$mocks/data/summary/episodes/silo/mapped/EpisodeSiloMappedMock.ts';
+import { ExtendedUsersResponseMock } from '$mocks/data/users/response/ExtendedUserSettingsResponseMock.ts';
+import { server } from '$mocks/server.ts';
+import { renderStore, setAuthorization } from '$test/beds/store/renderStore.ts';
+import { waitForValue } from '$test/readable/waitForValue.ts';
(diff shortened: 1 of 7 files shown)


seferturan (approved):
Applied some fixups to also use `useSpoilerFreeEpisodeTitle` in some other places 👌

PR #3015, opened 2026-07-27
Import: anchor date-only import values to noon UTC

### LINKED ISSUE #2964: default timezone results in inaccurate watch dates when importing Letterboxd data
### Code of Conduct

https://app.trakt.tv/settings/data?mode=movie&source=letterboxd#import

Data exported from Letterboxd does not include full datetimes on items, but rather just the date.  When importing data from Letterboxd, Trakt interprets that date as being 12 AM UTC on that day.

This can result in an off-by-one day error when converting the date to the user's current timezone.

For example, if a movie's watched date is 2026-07-20 in the Letterboxd export data, Trakt will record the movie as having been watched on 2026-07-20 at 12 AM UTC.  If a user in Los Angeles then views their watch history in Trakt, it will show the movie as having been watched on 2026-07-19 at 5 PM.

Defaulting to the user's current timezone (or one they select) rather than UTC could help address this issue.  Some arbitrary time will still need to be chosen since Trakt requires a time to be associated.  Even with a correct timezone, 12 AM might not be the best default since it means the off-by-one issue could reoccur if the user travels even one timezone over.  Something like 12 PM might be a bit more resilient in that the date will still be accurate when roaming up to ~12 timezones over in either direction.

I imagine this issue could be present when importing from other data sources as well, depending on whether they use full datetimes and whether they include a timezone/use UTC.  I can only speak from experience for Letterboxd data, however.

(Personal take, feel free to ignore) As a side note, I'm not a huge fan of watch dates being timezoned in the first place.  If I watched a movie at 8 PM in one timezone, I'm more interested in seeing it displayed as 8 PM no matter where I am.  One possible way to cater to this preference could be to allow the user to lock-in a set timezone for their account rather than always presenting the watch date in their device's current timezone.

### Additional information

_No response_

rectifyer:
Fixes #2964

# Problem

Letterboxd (and IMDb) exports carry date-only watch/rating dates like `2026-07-20`. Every import parser funnels these through `toISOString`, where `new Date('2026-07-20')` anchors the value to UTC midnight (`...T00:00:00.000Z`). Viewed from any negative-offset timezone (e.g. Los Angeles), midnight UTC rolls back to the previous calendar day, so a movie logged on 2026-07-20 shows up as watched on 2026-07-19.

# Fix

Anchor date-only imported values to noon UTC instead of midnight, centrally in `toISOString`. Noon UTC keeps the calendar day correct across the full ±12h span of world timezones (and roaming) — the resilient default the reporter suggested. Values that already carry a time component (Trakt CSV/JSON, TV Time exports) are detected and passed through untouched, so no other import source changes behavior.

Because the fix lives in the shared helper, it also corrects the same latent off-by-one in the IMDb parser (`Date Rated` is date-only too).

# Changes

- `toISOString.ts` — noon-anchoring for date-only values, with a doc comment explaining the timezone rationale.
- `toISOString.spec.ts` — new colocated spec covering noon-anchoring, LA/Tokyo calendar-day stability, full-datetime passthrough, and invalid inputs.
- `LetterboxdParser.spec.ts` / `ImdbParser.spec.ts` — updated expected `watched_at` / `rated_at` from midnight to noon.

# Notes

The reporter's secondary idea — a user-selectable fixed account timezone for display — is a larger, separate feature and is intentionally out of scope here. This change resolves the concrete off-by-one import bug. All 206 import-parser tests pass; changed files are `deno fmt`-clean.

### DIFF
--- a/projects/client/src/lib/sections/settings/import/parsers/ImdbParser.spec.ts
+++ b/projects/client/src/lib/sections/settings/import/parsers/ImdbParser.spec.ts
@@ -144,8 +144,8 @@ describe('ImdbParser', () => {
 
       const result = await ImdbParser.parse([dummyFile]);
 
-      expect(result[0].rated_at).toBe(new Date('2024-09-02').toISOString());
-      expect(result[1].watched_at).toBe(new Date('2024-09-02').toISOString());
+      expect(result[0].rated_at).toBe('2024-09-02T12:00:00.000Z');
+      expect(result[1].watched_at).toBe('2024-09-02T12:00:00.000Z');
     });
 
     it('handles missing Date Rated gracefully', async () => {
--- a/projects/client/src/lib/sections/settings/import/parsers/ImdbParser.ts
+++ b/projects/client/src/lib/sections/settings/import/parsers/ImdbParser.ts
@@ -2,7 +2,7 @@ import type { UniversalImportItem } from '../ImportTypes.ts';
 import type { FileParser } from './ParserInterface.ts';
 import { isValidItem } from './utils/isValidItem.ts';
 import { parseCsvFile } from './utils/parseCsvFile.ts';
-import { toISOString } from './utils/toISOString.ts';
+import { toImportISOString } from './utils/toImportISOString.ts';
 
 type ImdbRatingsRow = {
   Const?: string;
@@ -39,7 +39,7 @@ function parseRatingsRow(row: ImdbRatingsRow): UniversalImportItem[] {
 
   const type = toImportType(row['Title Type']);
   const rating = row['Your Rating'] ? Number(row['Your Rating']) : undefined;
-  const ratedAt = toISOString(row['Date Rated']);
+  const ratedAt = toImportISOString(row['Date Rated']);
 
   const ratingItem: UniversalImportItem = {
     action: 'ratings',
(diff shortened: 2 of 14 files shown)


danieltaylor:
Hi!  Thanks for tackling this issue.  As you mention, switching to noon UTC will correct the issue for anyone in a negative-offset timezone.  There are a few timezones in the UTC+12 – UTC+14 ranges however, and those would now have their watched dates pushed into the next day.

According to Wikipedia, "no inhabited territory lies within [the UTC-12] time zone offset," so we could safely bump the default up to 1 PM UTC, thus including the UTC+12 zones (such as most of New Zealand, Fiji) within the correct date.

1 PM UTC would cover the most inhabited timezones possible, however it would still leave out a few areas.  If we are to ensure no one gets left out, we would likely need to account for the user's timezone in some way.

seferturan (approved):
Applied some fixups:

- Dropped the `Z`, so date-only values anchor to noon **local** instead of noon UTC.
  - @danieltaylor is right that noon UTC pushes UTC+12 to UTC+14 into the next day. So went with @danieltaylor's suggestion: account for the user's timezone. Anchoring noon local is correct in every zone from -12 to +14, and keeps the ~12h roaming slack.
- Test was asserting LA and Tokyo, both on the safe side of the boundary, so it couldn't catch this. Swapped for a local-timezone invariant.
- Also renamed `toISOString` to `toImportISOString` in a separate commit. It shadowed the built-in while behaving differently, and TvTimeNative/TvTimeGdpr call the real one a line away from the helper.

PR #3022, opened 2026-07-28
feat(settings): add official apps catalog

kevincador:
## Summary

Adds an official Trakt app catalog to the Applications settings page while keeping connected app management focused on revocable community apps.

## Changes

- Display official apps in Apple, Android, and Web tabs.
- Use normalized local app icons and open store links in new tabs.
- Add a Public API Docs row to the applications card.
- Add focused coverage for the catalog and links.

## Screenshots

Screenshot 2026-07-28 at 16 29 56
Screenshot 2026-07-28 at 16 30 23

## Risk

Low. The catalog is static, so store links and local icons will require manual updates if an app changes.


seferturan:
I don't think this belongs in settings. If we do want something like I'd say either in the `about` page, or a separate `apps` page.

PR #3027, opened 2026-07-28
Calendar: episode type filter

### LINKED ISSUE #2948: Calendar filters
### Code of Conduct

The ability to filter the calendar for season/series finales is a “can’t live without “ issue.
I cannot be the only binge watcher using Trakt!

Expand Calendar filtering in Web V3 to support additional filters like premieres, finales, and more granular release types. We should also introduce an "All" vs "My Services" mode (e.g. all episode releases, not just followed shows), with advanced filters available in "All" to improve discovery without cluttering the default view.

_No response_

### Additional context

_No response_

rectifyer:
Closes #2948

# Summary

Adds an episode type filter to the calendar, so the feed can be narrowed to premieres and finales. This is the v0 discussed on the issue: episode roles only, with the broader "All vs My Services" mode left for a follow up.

The simple filter is a single select offering Premieres, Finales and Series Premiere. The advanced filter is a multi select across all six roles, with `+` / `-` include and exclude support like the status filter. Including a role narrows the feed to matching episodes; excluding one removes those episodes and leaves everything else, so in the merged media view a movie survives an excluded role because it carries no role of its own.

# Screenshots

### Simple Filtering
image

### Advanced Filtering
image

# Client side filtering

The calendar endpoints take no episode role parameter, so this filter is applied to the entries the calendar already fetched rather than being sent upstream. `Filter` gained an `isClientSide` flag that keeps the key out of `filterMap`, which means it never reaches a request that has no parameter for it and never lands in the query dependencies — toggling it re-filters what is already in hand instead of refetching.

`useCalendar` takes the selection as an `Observable` rather than a plain value, so a change does not re-instantiate the hook and rebuild the whole `QueryObserver` chain. The narrowing runs ahead of the intl overlay so we do not fetch translations for entries that are about to be dropped.

One wrinkle worth knowing about: the calendar requests `group=day`, and a day that holds both a premiere and a finale arrives collapsed into a single `full_season` card with its individual episodes dropped. `multiple_episodes` cards keep theirs and are matched exactly. For `full_season` all that survives is that the span opens and closes a season, so those cards answer to any premiere or finale role rather than being hidden from every one of them. The trade is a `full_season` card showing up under Series Finale for a show that is not actually ending, which is preferable to a genuine series finale being hidden.

# Filter scoping

Episode roles only mean something on the calendar, and only when shows are in view, so `Filter` also gained two optional scoping fields rather than the filter being added globally.

`surfaces` restricts a filter to pages that opt in via `filterSurface` on `NavbarStateSetter`, and `modes` restricts it to discover modes. Omitting both keeps a filter global, which is what every existing filter does. The episode type filter declares `surfaces: ['calendar']` and `modes: ['show', 'media']`, so it stays out of the watchlist, lists, history and discover panels, and disappears when the calendar is switched to movies.

The gating lives in a new `useVisibleFilters`, deliberately separate from `useFilter`. Reading the discover mode requires the discover context, and `useFilter` is called by roughly twenty list sections that only want `filterMap` — none of them should take on that dependency. `useFilter().filters` therefore stays the complete set, because the URL and stored filter plumbing has to recognise every key regardless of which page it came from.

`processFilterParams` now skips surface restricted keys so that "save as default" cannot persist a calendar only filter and replay it onto every other filterable page's URL.

# Multi select include styling

The `+` state now renders its label in `--purple-500`, mirroring the existing `--red-500` treatment on `-`. Both tokens sit in the same lightness band, so the two states read with equivalent contrast, and the include colour matches the chip that sets it. This is a shared component, so it also applies to the other advanced multi select filters, and to check style multi selects where a selected row is still "included".

# Known gap

Switching the calendar to movies hides the control but leaves an existing `episode_types` value in the URL, where it still counts toward the active filter badge. It has no effect on results, and "Reset all" clears it. Stripping now invalid filter params on mode change belongs in `useDiscover.onModeChange`, which is shared well beyond the calendar, so it is left for a separate change.

# Testing

Unit coverage for the role matching predicate, the new filter definition, the visibility predicate and the stored filter exclusion. Full client suite passes with no new failures.


vladjerca:
This implementation does not follow the discussed behavior that the team agreed on... Episode type is not a globally applicable filter.

PR #3039, opened 2026-07-28
Calendar: episode type filter

### LINKED ISSUE #2948: Calendar filters

rectifyer:
Closes #2948

# Summary

Adds an episode type filter to the upcoming feeds, so the calendar can be narrowed to premieres or finales. This is the v0 discussed on the issue: episode roles only, with the broader "All vs My Services" mode left for a follow up.

It is a three way toggle rather than an entry in the filters panel: **All** (default), **Premieres**, **Finales**. The control sits beside the title on both surfaces that show upcoming releases, so it reads as part of the feed rather than something buried behind the filter drawer.

# Screenshots

## Home > Calendar > All
image

## Home > Calendar > Premieres
image

## Home > Calendar > Finales
image

## Calendar > All
image

## Calendar > Premieres
image

## Calendar > Finales
image

## Calendar > Movies
image

# How it works

The toggle is a normal entry in the shared toggler registry, so it is the same component and behaviour as the Progress toggle on a profile. It is backed by `useToggler('episodeType')` for persistence and mirrored onto an `episode_type` URL parameter through `GlobalParameterSetter`, the same way the discover mode toggle works. The parameter wins whenever it holds a known value, which keeps a filtered calendar shareable, and home and the calendar page read the same toggler so a choice made on one carries to the other.

Filtering happens on the client. The calendar endpoints accept no episode role parameter, so `matchesEpisodeTypeFilter` narrows the entries that already arrived rather than anything being sent upstream. `useCalendar` and `useUpcomingItems` both take the selection as an `Observable`, so changing it re-filters what is in hand instead of re-instantiating the hook and rebuilding the whole `QueryObserver` chain. The narrowing runs ahead of the intl overlay so we do not fetch translations for entries about to be dropped, and on home it runs ahead of the row limit so the row fills with matches rather than with whatever survived the first page.

Premieres covers series, season and mid-season premieres; Finales covers series, season and mid-season finales.

# Grouped days

The upcoming feeds request `group=day`. A day holding several episodes of one show arrives as a single `multiple_episodes` card that still carries its episodes, so those are matched exactly. A day holding both a premiere and a finale arrives as a `full_season` card with the individual episodes dropped, so all that survives is that the span opens and closes a season. Those cards therefore answer to both filters rather than being hidden from each of them, which is the right trade for the binge watching case the issue describes: a full season drop genuinely is both.

# Infinite scroll

The calendar stops loading further periods once enough of them come back empty. Filtering after the fetch would make a narrow selection empty almost every window, so the calendar would stop scrolling after five of them even though matches exist further out. `useCalendar` now exposes `hasUpstreamItems`, taken before the filter runs, and the accumulator's emptiness check accepts it as an override. The heuristic answers "is the feed exhausted", not "did the filter hide everything".

# Movies

A movie carries no episode role, so the toggle hides itself when the discover mode is movies and the selection is forced back to All while it is hidden. Without the second half, switching to Movies while on Premieres would leave an empty feed with no visible control to undo it.

# Icons

Two new icons, both taken from Material Symbols so they carry the same weight and grid alignment as the rest of the set: a four point sparkle for Premieres, and the checkered flag (`sports_score`) for Finales.

# Also in here

The `+` state on advanced multi select filters now renders its label in `--purple-500`, mirroring the existing `--red-500` treatment on `-`. Both tokens sit in the same lightness band so the two states read with equivalent contrast, and the include colour matches the chip that sets it. This is a shared component, so it applies to every advanced multi select filter.

# Testing

Unit coverage for the role matching, including grouped days, full season cards and the movie case, plus both directions of the accumulator override. Full client suite passes with no new failures.

### DIFF
--- a/projects/client/i18n/meta/en.json
+++ b/projects/client/i18n/meta/en.json
@@ -1560,6 +1560,30 @@
       "default": "Status",
       "description": "Header for the status section."
     },
+    "button_text_episode_type_all": {
+      "default": "All",
+      "description": "Text for the toggle that shows every upcoming release, rather than only premieres or only finales. This should be as short as possible."
+    },
+    "button_label_episode_type_all": {
+      "default": "Show all releases",
+      "description": "Accessible label for the toggle that shows every upcoming release, rather than only premieres or only finales."
+    },
+    "button_text_episode_type_premieres": {
+      "default": "Premieres",
+      "description": "Text for the toggle that shows only premiere episodes: series, season and mid-season premieres. This should be as short as possible."
+    },
+    "button_label_episode_type_premieres": {
+      "default": "Show only premieres",
+      "description": "Accessible label for the toggle that shows only premiere episodes: series, season and mid-season premieres."
+    },
+    "button_text_episode_type_finales": {
+      "default": "Finales",
+      "description": "Text for the toggle that shows only finale episodes: series, season and mid-season finales. This should be as short as possible."
+    },
+    "button_label_episode_type_finales": {
+      "default": "Show only finales",
+      "description": "Accessible label for the toggle that shows only finale episodes: series, season and mid-season finales."
+    },
     "translated_value_status_released": {
       "default": "Released",
       "description": "Status value for released media items. The value is sent back from the Trakt API."
--- a/projects/client/src/lib/components/icons/FinaleIcon.svelte
+++ b/projects/client/src/lib/components/icons/FinaleIcon.svelte
@@ -1,0 +1,11 @@
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  height="24px"
+  viewBox="0 -960 960 960"
+  width="24px"
+  fill="currentColor"
+>
+  <path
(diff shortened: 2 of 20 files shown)


rectifyer:
@ElMagnea Could use your input on the premieres and finale icons. Premieries needs to be something else since this one is too close to "discover". The finale might be ok, but perhaps just a cleaner "checkered flag". Open to completely new ideas too.

ElMagnea:
> @ElMagnea Could use your input on the premieres and finale icons. Premieries needs to be something else since this one is too close to "discover". The finale might be ok, but perhaps just a cleaner "checkered flag". Open to completely new ideas too.

Clapperboard or Rocket and Checkerboard? 

image

rectifyer:
## Updated Icons

image

michaldrabik:
Personally not a fan of using TV icon for "All" simply because in every other place in the app TV icon is for "TV Shows only" filters etc etc.

That first one with TV/Movie Reel works better imo -> or just completely something new that would not collide?

seferturan:
> Personally not a fan of using TV icon for "All" simply because in every other place in the app TV icon is for "TV Shows only" filters etc etc.
> 
> That first one with TV/Movie Reel works better imo -> or just completely something new that would not collide?

Agreed. I'll change it to match the icon of the currently selected discover mode. I.e. tv/movie reel if media, tv if show, reel if movie

seferturan (approved):
Applied some fixups:

- Used the current discover modes icon for `all`. We can tweak as a follow up if need be.
- Added the selected option as a meta info label under the title.
- For now, I made it behave as the other list toggles (without search params). This is something we should tackle as a separate thing and fix generically for all toggles on all pages.

PR #3041, opened 2026-07-28
Plex Sync: available to free members with limits

rectifyer:
# Summary

Opens Plex Sync to free members, matching the server side update that lifted the VIP-only restriction, and reworks the settings page around an add-a-server flow: connect your Plex account, add a server, and manage its libraries from a drawer. Free accounts sync 1 Plex server once a day; VIP syncs unlimited servers refreshed hourly. The Plex webhook (real-time scrobbler) tab remains VIP only.

# How it works

- The Connection card is unchanged: connect or disconnect the Plex account.
- **Synced Servers** lists one row per syncing server: name, green Syncing badge, its selected libraries, a per-server Sync Now (VIP only), and a Manage button. An empty state prompts adding the first server.
- **Add Server** is a standard select dropdown listing the account's not-yet-added servers. Picking one adds its row and opens the manage drawer. A free member who already has a server syncing gets an inline VIP upsell instead.
- **Manage** opens a side drawer with the Sync as account picker and per-library switches. Changes are staged locally and applied in a single request via the check action in the drawer header (same pattern as the Favorites drawer), with a discard-changes confirmation when closing with unsaved edits.
- **Plex Sync Settings** (per-media-type toggles) stays global and only renders once at least one server is actually syncing.
- The history section is now titled **Data Syncs** to match the streaming settings page: same subtitle treatment, a Plex logo badge on each sync row, and the summary banner showing total sync count and most recent sync.

# Implementation notes

- The tier is derived from `user.isVip` client side (free = 1 server, VIP = unlimited — the same rule the API applies). Saving a free member's selection sends only the current server's libraries, mirroring the API's clamp so the selection moves with the server being configured instead of being silently dropped. The `sync.server_limit` field in the `plex_settings` response isn't in the `@trakt/api` contract yet; once published, the client can consume it directly.
- A freshly added server is session-local until its first library selection is applied — the API derives a free member's allowed server from the library selection, so nothing is persisted for an empty server.
- The Data Syncs summary banner was extracted into a shared `DataSyncsBanner` used by both the streaming and Plex sections, and paginated page meta gained an optional `itemCount` read from the `x-pagination-item-count` header to supply the total count.
- Obsolete VIP-only upsell i18n keys were replaced with new keys (free-limits banner, synced servers, manage drawer, add-server upsell) so stale locale translations can't keep claiming sync is VIP only.

# Screenshots

### Adding a new server
image

### Managing libraries
image

### Plex sync settings
image


kevincador:
@rectifyer looking good. Only question I have, looking at the screenshot, is what does the "Sync Now" button do on the free tier? Do we have limits there to how many times a sync for a free user can be run manually? Some kind of abuse/automation prevention?

rectifyer:
@kevincador Good catch, the sync now button is only available for VIPs now.

StefanFriptu (commented):
Image

Waiting on fix approval regarding empty accounts on PMS

seferturan (approved):
Applied some fixups:
- `saveChanges` could hang forever. It awaited a filtered `user` stream, so an unresolved user query left Apply stuck disabled.
- Failed saves were silent, now surfaced in the drawer.
- Restored the account load error state in the manage drawer, it got dropped in the restructure.
- Adding a server then closing the drawer without applying left a phantom row that ate the free tier slot until reload.
- Apply with no libraries selected silently removed the server. Now disabled, with a hint saying why.
- Dropped `itemCount` from `extractPageMeta` / `Paginatable`. We plan to drop `x pagination-item-count`, so no point baking it into the shared model. Sync count banner goes with it. If we want the counts later on, should be in the API respose.
- Only one upsell at a time now. Top one shows pre-connection only, the multiple servers one sits between the title and the server list.
- Add server is a popup menu with a plus icon instead of a select. A select for an action felt off.
- Remove server added. Trash button next to the server name in the manage drawer, with a confirmation.
- Drawer footer is full width cancel/apply, like the other settings drawers.
- Left a `fixme` in there for when we update the api package.

Examples with those changes:

Image

Image

Image

Image

PR #3053, opened 2026-07-30
feat(tags): reverse poster tag contrast in the dark theme

ElMagnea:
## Visuals

Poster indicator tags now render as a light fill with a charcoal glyph in the dark theme, so they separate from cover art instead of sinking into it.

Screenshot 2026-07-30 at 12 52 10

The light theme deliberately keeps the original light-on-dark scheme. These tags straddle the cover's bottom edge, so the overhanging half sits on the page — and a light fill there would match `--color-background` exactly and lose its boundary entirely. The pair is theme-mapped in both mixins rather than pinned.

Screenshot 2026-07-30 at 12 52 23

## Shape and sizing

Markers and the summary poster pills are aligned:

- **Pill radius** drops to `--border-radius-s`. `--border-radius-m` was being capped at half the pill's height and rounding off into a stadium, so it never rendered as specified.
- **Summary poster pills** take the same radius. Scoped to `SummaryPosterTags` rather than the shared `TagContent` (used by every tag in the app), and rather than `SummaryPoster`, whose tags slot carries an IMDb link on person pages.
- **Glyph heights normalised.** Each glyph fills a different share of its own viewBox, so a shared box rendered them at different sizes — rewatching was coming out at half the height of the rest. Rewatching stops short of full parity at `1.6`, since `2` would make it the widest glyph in a pill sized to the NEW / PREMIERE cover tags.
- **Pill scales up 15%** for presence on the cover, and is sized against those same cover tags: 20.7px tall with 6.9px inline padding against their 20px / 6px.
- **Footer clearance.** `--height-card-footer-sm` grows 28px → 36px and the transparent-card footer gains top padding, so the info row clears the straddling tag's overhang. Every poster list height derives from that token, so Related / Recommended / Smart list heights grow with it.

## Also

Removes the dead `--color-*-watch-count-tag` pair, which no component consumed, and repoints the design-system colour page at the tokens that are actually live.

## Testing

`deno fmt --check` clean · `svelte-check` 0 errors / 0 warnings across 9179 files · 2515 unit tests passing. Verified in both themes on poster cards and summary pages.

## Known follow-up

The partially-watched marker's half-fill gradient uses `to right`, so in RTL locales the filled half points the wrong way. It encodes progress rather than decoration, so per `components.md` it should mirror — a one-line fix with `--rtl-sign`, kept out of this colour-only change.

### DIFF
--- a/projects/client/src/lib/components/card/CardFooter.svelte
+++ b/projects/client/src/lib/components/card/CardFooter.svelte
@@ -31,7 +31,7 @@
 <style>
   :global(.trakt-card-transparent) {
     .trakt-card-footer {
-      padding: var(--ni-8) 0 0;
+      padding: var(--ni-10) 0 0;
     }
   }
 
--- a/projects/client/src/lib/components/media/tags/DroppedTag.svelte
+++ b/projects/client/src/lib/components/media/tags/DroppedTag.svelte
@@ -21,7 +21,5 @@
     {/snippet}
   </StemLinkTag>
 {:else}
-  <IndicatorTag>
-    <DropIcon />
-  </IndicatorTag>
+  <IndicatorTag indicator="dropped" />
 {/if}
--- a/projects/client/src/lib/components/media/tags/IndicatorTag.svelte
+++ b/projects/client/src/lib/components/media/tags/IndicatorTag.svelte
@@ -1,30 +1,69 @@
 <script lang="ts">
+  import BookmarkIcon from "$lib/components/icons/BookmarkIcon.svelte";
+  import DropIcon from "$lib/components/icons/DropIcon.svelte";
+  import FastRewindIcon from "$lib/components/icons/FastRewindIcon.svelte";
+  import TrackIcon from "$lib/components/icons/TrackIcon.svelte";
   import StemTag from "$lib/components/tags/StemTag.svelte";
 
   const {
-    children,
     variant = "full",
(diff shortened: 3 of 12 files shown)


seferturan (approved):

michaldrabik:
@ElMagnea @seferturan 

Have you guys also increase the size of tags by accident? They look really big and bulky now I gotta say when looking at it on real device :D 

Maybe overshot a bit - color update should be enough?

image

image

ElMagnea:
Good eye @michaldrabik — yes, that was accidental, and you're right that the colour update alone was the intent.

There were two separate scales in play and only one was meant to move. `--glyph-scale` corrects the icon inside the tag: the double-check fills only ~80% of its viewBox where the bookmark fills 100%, so it rendered noticeably shorter sitting next to it. That one's doing real work. But the tag also carried `scale: 1.15` on the pill itself, which blows up the background, padding and radius along with the glyph — 22×18px → 25.3×20.7px. That's the bulk you're seeing.

Fixed in #3101: the pill goes back to the size it's always been, the glyph corrections stay, and as a bonus the tag's corner lands back on exactly the 6px the summary poster pills use instead of 6.9px.

While in there I also derived the footer clearance from the tag's actual height rather than the hardcoded padding it had. On small cards (Related / Recommended) the info row was down to ~2px below the tag — that spacing had drifted every time the tag got resized, because the overhang and the padding compensating for it lived in different files with nothing tying them together.

PR #3057, opened 2026-07-30
feat(settings): make api application creation vip-only

vladjerca:
Makes API application creation VIP-only, and folds the new upsell surface into a shared settings component.

## Behavior

- Free users see a VIP upsell instead of the create-application form on `/settings/apps/api/new`, and an upsell card above the (still readable) application list on `/settings/apps/api`.
- The create button in the list header only renders for VIP.
- VIP behavior is unchanged.

## Shared upsell component

`PlexVipUpsell` and the new API-application gate were near-identical (same markup roles, same CSS values), so both now go through `settings/_internal/SettingsVipUpsell.svelte`, with the icon passed as a snippet.

Plex settings picks up the newer look as a side effect:

- column -> row layout, with `for-mobile` stacking
- divider removed
- hand-rolled `purple-500 14%` background -> `--background-vip-elevated-card` + `--color-vip-border-accent`
- adds the vip card box-shadow

Also dropped a dead `.upsell-description` rule - the `secondary` typography class already sets `--color-text-secondary` and outranks it on specificity.

## Verification

- `deno task check` - 0 errors
- `deno fmt --check` - clean
- eslint - no findings on touched files
- `deno task test:unit` does not run in this worktree: `projects/client/node_modules/vitest` symlinks into a sibling clone, so vitest fails during its own config bootstrap before any spec is parsed. Unrelated to this change; CI runs it properly. No spec covers the touched components.


antwanchild:
@vladjerca 
Is there a bug?  I thought non VIP were still allowed 1 app?  My 1 and only app got deleted and no way to created a new one

vladjerca:
@antwanchild free users are allowed to **connect** 1 community app.

YozoraXCII:
> @antwanchild free users are allowed to **connect** 1 community app.

So free users can no longer have their own app, and instead can only connect to a community app presumably made by a VIP user?

kevincador:
Thanks for the feedback. This PR has been merged and is considered resolved. The recent discussion concerns the underlying business decision rather than the implementation, so it’s no longer productive or relevant to this PR review. We’re locking this conversation to keep the PR focused and constructive.

PR #3058, opened 2026-07-30
Season: use season titles if available

rectifyer:
# Summary

Some shows have custom season titles (e.g. American Horror Story's "Murder House", "Asylum", "Coven"). The API returns a title for every season, but only ones that differ from the generic "Season {number}" / "Specials" defaults are real custom titles, so the mapper normalizes generic titles to null and everything downstream treats `season.title` as "has a custom title".

# Changes

- `Season` model gets a nullable `title`, populated by `mapToSeason` only when the API title isn't a generic default.
- `SeasonDropdown` labels become `1: Murder House` when a custom title exists (plain `1` otherwise, `Specials` for season 0). This covers the summary page dropdown, the seasons drawer badge, and the episodes tab in the episode drawer, since they share the component.
- Season poster cards (`SeasonItem`) show the custom title as a second line under the always-bold `Season 1` line. The selected season's number line uses the primary text color; footer text is top-aligned so title-less cards line up.
- The seasons drawer strip and mobile poster list switch to the taller card footer only when the show actually has custom titles, so shows without them keep the compact layout.
- New i18n key `text_season_number_with_title` (`{number}: {title}`) so the separator is translatable.
- Silo mapped season mock updated with `title: null`.

# Screenshot

image

### DIFF
--- a/projects/client/i18n/meta/en.json
+++ b/projects/client/i18n/meta/en.json
@@ -1502,6 +1502,18 @@
         }
       }
     },
+    "text_season_number_with_title": {
+      "default": "{number} - {title}",
+      "description": "Compact dropdown label combining a season number with its custom season title, e.g. '1: Murder House'.",
+      "variables": {
+        "number": {
+          "type": "number"
+        },
+        "title": {
+          "type": "string"
+        }
+      }
+    },
     "text_season_specials": {
       "default": "Specials",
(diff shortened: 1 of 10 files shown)


ElMagnea (commented):
Makes perfect sense to me

seferturan (approved):

PR #3069, opened 2026-08-03
fix: premiere/finale tags for coalesced episodes

seferturan:
## ~~⚠️ Drafted ⚠️~~
~~Drafted to deal with the 'premier' clash~~

## 🎶 Notes 🎶

- ~~Adds a `New Season` tag to calendar and releases lists, and their drilled-down pages~~
- The root issue was that `premiere` tag was never shown for coalesced episodes, e.g. when ep 1 and 2 would released on the same day for a new season.
- Also adds `New` variants to the premier/finale tags.
  - Did not go for separate `new` and `premiere` tags since on smaller devices it won't fit.
  
## 👀 Example 👀
Example with mock data:

Screenshot 2026-08-25 at 15 28 41


projects/client/src/lib/components/episode/tags/EpisodeStatusTag.svelte
```diff
@@ -20,59 +23,81 @@
...
+    display: contents;
 
-{#if status === "new"}
-  {@render tag(i18n.newText())}
-{/if}
+    :global(.episode-status-tag),
+    :global(.episode-status-tag .trakt-tag) {
+      min-width: 0;
+    }
+  }
```
rectifyer: **Convention: consumer restyles shared tag components via `:global()`.**

This wrapper div exists only to inject `min-width: 0` into `TextTag`/`StemTag` from the outside, which is the pattern `components.md` forbids under "Restyling a Shared Component" ("If the look you need does not exist, add it to the component so every caller gets it").

Truncation-friendly tags are a general need, so consider supporting shrinking in `TextTag`/`StemTag` themselves (a prop or a documented CSS variable hook). That would also let you delete this wrapper, the `display: contents` trick, and the `classList="episode-status-tag"` plumbing.

projects/client/src/lib/components/episode/getEpisodeStatus.ts
```diff
@@ -1,47 +1,95 @@
...
+  episodes?: CoalescedEpisodes;
 };
 
+const PREMIERE_TYPES: ReadonlySet<EpisodeType> = new Set(
+  Object.values<EpisodeType>(EpisodePremiereType),
+);
+
+const FINALE_TYPES: ReadonlySet<EpisodeType> = new Set(
+  Object.values<EpisodeType>(EpisodeFinaleType),
+);
```
rectifyer, line 23: **Drift risk: two competing definitions of premiere/finale.**

These sets are derived from the enums and include the mid-season variants, while `coalesceEpisodes` decides `full_season` from narrower inline lists (`['season_premiere', 'series_premiere']` / `['season_finale', 'series_finale']`). That difference is intentional today (a mid-season premiere day should not coalesce to `full_season`), but the two definitions live far apart and can drift silently.

Suggestion: share named constants between the two, or at least leave a comment in `coalesceEpisodes` stating the narrower set is deliberate.

projects/client/src/lib/components/episode/CoalescedEpisodes.ts
```diff
@@ -0,0 +1,5 @@
+import type { EpisodeEntry } from '$lib/requests/models/EpisodeEntry.ts';
+
+export type CoalescedEpisodes =
+  | ReadonlyArray<Pick<EpisodeEntry, 'type'>>
+  | Nil;
```
rectifyer, line 5: **Typing nit: double optionality.**

`| Nil` is baked into the alias while every usage site also declares the prop/option as `episodes?: CoalescedEpisodes`, so the nullability is expressed twice.

Suggest dropping `| Nil` here:

```ts
export type CoalescedEpisodes = ReadonlyArray<Pick<EpisodeEntry, 'type'>>;
```

and widening the one internal consumer accordingly (`resolveEpisodeType(type, episodes?: CoalescedEpisodes)`); it already handles the nullish case with `?? []`.

rectifyer (changes requested):
Nice fix overall: resolving the milestone from the coalesced children is clean, the combined `new-premiere` / `new-finale` statuses are a good call for narrow screens, and test coverage is thorough (coalesced days, future release dates, full-season premiere preference, the mid-season gate falling back to New).

Requesting changes for one convention issue (restyling shared tag components from a consumer via `:global()`), plus two smaller nits inline. Everything else checked out: `status` in `EpisodeItem` is only used for truthiness so the new combined values are safe, `episodes` was already nullish on `EpisodeEntrySchema` so attaching it to `full_season` entries is safe for existing consumers, and the merged `upcoming`/`calendar` block preserves the previous tag order.

One behavior change worth a line in the PR description: the mid-season gate (`isLatestAired === false`) previously suppressed the tag entirely; now a recent release falls back to "New".

projects/client/src/lib/components/episode/tags/EpisodeStatusTag.svelte
seferturan: Good catch 👌

Moved `min-width: 0` into `TextTag` and `StemTag` themselves, so every caller gets shrinkable tags. Wrapper div, `display: contents` and the `classList` plumbing are all gone now.

`ProgressTag` was already doing this internally, so this just makes it consistent.

projects/client/src/lib/components/episode/getEpisodeStatus.ts
seferturan: Half agree. Not sharing constants across the layers though, `lib/requests` shouldn't reach into `lib/components/episode` for that.

Went for the smaller fix instead: `coalesceEpisodes` was matching on raw string literals (`['season_premiere', 'series_premiere']`), which is `string[]`, so `.includes(ep.type)` was type checking against nothing. Rename an enum value and it silently stops matching.

Now it's `FULL_SEASON_PREMIERE_TYPES` / `FULL_SEASON_FINALE_TYPES` built from `EpisodePremiereType` / `EpisodeFinaleType`. Names carry why the set is narrower, and it's actually typed now.

projects/client/src/lib/components/episode/CoalescedEpisodes.ts
seferturan: No, these aren't the same optionality. `EpisodeEntry.episodes` is `.nullish()` (`EpisodeEntry.ts:32`), so it comes through as `T[] | null | undefined`. The `| Nil` in the alias absorbs the `null`, the `?` at the call site absorbs the `undefined`.

Tried your version, it doesn't compile:

```
Type '...[] | null | undefined' is not assignable to type 'CoalescedEpisodes | undefined'.
  Type 'null' is not assignable to type 'CoalescedEpisodes | undefined'.
```

Could push the `| Nil` out to the call sites instead, but that's three spots rather than one. Leaving as is 👌

PR #3071, opened 2026-08-03
feat: invert lists cards posters

michaldrabik:
- this was annyoing me for a bit, maybe it makes sense to change
- for lists cards the posters were layed out from right to left -> It might make more sense to change it to natural order which also matches what you see if you go into list items details

NOW:
image
image

AFTER: 

image
image

### DIFF
--- a/projects/client/src/lib/sections/lists/components/list-summary/_internal/ListPosters.svelte
+++ b/projects/client/src/lib/sections/lists/components/list-summary/_internal/ListPosters.svelte
@@ -80,6 +80,7 @@
 
     position: absolute;
     inset-inline-start: calc(var(--poster-offset) * var(--poster-index));
+    z-index: calc(var(--poster-count) - var(--poster-index));
 
     height: var(--poster-height);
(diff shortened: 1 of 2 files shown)


ElMagnea:
@michaldrabik my mental analogy of this is a deck of cards and the dealer sliding on the table. It goes left to write. We're also currently showing the top 8 items in that list, so altering the order of those items will be reflected in the list preview posters. I don't see anything wrong with this so unless it's just a left to right, right to left personal preference I think we can close this and leave it as is. Tell me if i'm missing something.

michaldrabik:
@ElMagnea Taking deck of cards analogy the top card in my "Before" example should be Project Hail Mary, yet its FROM ;) Then its just direction of the slide but the top card would remain PHM as in L2 list #1 position.

But if you still think its not worth chasing just close it since its mostly QoL change.

seferturan:
Yeah, reads weird to me also 😅 And this ordering is fine for me also, since the top card is the one that was added to the list later.

PR #3074, opened 2026-08-03
List: retain sorting params when mode is switched

### LINKED ISSUE #3046: Lists: Mode Toggles Reset Sorting
### Code of Conduct

v3 Desktop, List Detail Pages

- Change the sorting
- Use the mode toggle at the top of the page (not the ones from the filter drawer), e.g. switch to movies only
- The sorting is reset, as is reflected by the URL (only the `mode` query param remains)

...the sorting should remain, as it does when using the mode filters.

### Additional information

_No response_

rectifyer:
# Problem

On list detail pages, switching the mode via the top-of-page toggles (e.g. movies only) dropped the active sorting: the resulting URL kept only the `mode` query param. The mode filters in the filter drawer preserved sorting correctly.

Fixes #3046

# Root cause

The mode toggle anchors get their hrefs rewritten by the `appendGlobalParameters` action, which is responsible for carrying the page-local `sort_by`/`sort_how` params onto same-path links. That rewrite only re-ran when one of its subscribed streams (`search`, `override`, `isEscaped`, `filterScopeStore`) emitted. Changing the sorting is a same-path navigation that touches only local params, none of the whitelisted global ones, so nothing emitted and the toggle's href stayed frozen at its mount-time value (`?mode=movie` with no sort params). Clicking the toggle then navigated to that stale URL, wiping the sort.

The filter drawer toggles were unaffected because `applyFilterToUrl` builds from the live `page.url` at click time.

# Fix

`createParameterContext` was already maintaining a `url` BehaviorSubject via `afterNavigate` but never exposed it (dead code, along with two other unused subjects which are now removed). This PR exposes it through `useParameters` and adds it to the action's `combineLatest`, so anchor hrefs are recomputed after every navigation. `applyParams` already reads the current location live, so it picks up the sort params on re-run.

# Testing

Added a regression spec: mount a `?mode=movie` anchor under a `mode` override context, navigate to a URL with `sort_by=rank&sort_how=asc`, and assert the rewritten href contains both the sort params and the mode.

### DIFF
--- a/projects/client/src/lib/features/parameters/GlobalParameterProvider.spec.ts
+++ b/projects/client/src/lib/features/parameters/GlobalParameterProvider.spec.ts
@@ -1,0 +1,27 @@
+import GlobalParameterHost from '$test/beds/parameters/GlobalParameterHost.svelte';
+import { render } from '@testing-library/svelte';
+import { beforeEach, describe, expect, it } from 'vitest';
+
+describe('GlobalParameterProvider', () => {
+  beforeEach(() => {
+    globalThis.window.history.pushState(
+      {},
+      '',
+      '/profile/userA?mode=movie&sort_by=rank&sort_how=asc',
+    );
+  });
+
+  it('should append its context parameters to anchors rendered under it', () => {
+    const { getByTestId } = render(GlobalParameterHost, {
+      props: { href: '?mode=movie', parameter: 'mode' },
+    });
+
+    const params = new URL(
+      (getByTestId('link') as HTMLAnchorElement).href,
+    ).searchParams;
+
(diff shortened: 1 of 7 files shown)


seferturan (approved):
Applied some fixups: 
- `url` now lives in the shared parameter context, published by `GlobalParameterProvider` from the `$effect` it already had, instead of a subject + `afterNavigate` per `Link`/`Toggle`/button host
- Exposed it read-only with a `setUrl` writer so only the provider can publish, mirroring the existing `search`/`update` split
- `applyParams` reads the URL off the emission instead of `globalThis.window.location`, so `url` is a real dependency rather than just a trigger

PR #3076, opened 2026-08-04
fix(stats): fix undercounted streaks from rewatched episodes

tysonkerridge:
Watched-show history was collapsing every rewatch of an episode down to a single date, so the streak feature was undercounting real activity — sometimes badly, since long-running streaks could be off by years for anyone who rewatches episodes.

This PR bundles the fix plus its knock-on effects:

- **Streak data fix**: watched-shows mapping now keeps every play date instead of only the most recent one per episode, fixing current streak, previous streak, dropped streaks, and active-day counts in the "Daily Streak Activity" drawer.
- **Toast regression fix**: the show-rating "binge" prompt was relying on that same collapsed data shape. Left alone, the data fix above would have made it misfire off of repeat plays of a single episode instead of genuine binge-watching across distinct episodes.
- **Display fix**: streak counts in the drawer now render as full numbers (`1,794`) instead of compact notation (`1.8K`), since streaks can legitimately run long enough for that compaction to hide real precision.

**Test plan**
- [x] `deno task test:unit` passes (or at least `currentUserWatchedShowsQuery.spec.ts` and `isShowRatingCandidate.spec.ts`)
- [x] Open the streak drawer for an account with rewatched episodes and confirm current/previous streak and active-day counts reflect real activity
- [x] Confirm streak numbers render in full (`1,794`) rather than compact (`1.8K`)
- [x] Confirm rewatching a single episode 3+ times in a week does not trigger the "rate this show" binge prompt

Before/After:
Screenshot 2026-08-04 at 12 00 07 pm Screenshot 2026-08-04 at 11 55 26 am

iOS (counts now match After above):
Screenshot 2026-08-03 at 11 14 36 pm


seferturan (approved):
I applied some fixups:

  - Simplified the mapper a bit: `watchedDates` now derives straight from `seasons` instead of building an intermediate list and stripping the extra field back off
  - `statCard` takes a `format` fn now (defaults to `toHumanNumber`) instead of a `string | number` union. Streak cards pass `toGroupedNumber`, days-active passes `toPercentage`. Keeps formatting in one place so a new card can't silently fall back to compact notation
  - Added a rewatch to `WatchedShowsMinimalResponseMock` so the query spec actually covers the fix. Verified it fails without the mapper change 👌
  - Few more mapper tests: rewatch dates across seasons, `playsPerSeason` staying an episode count, `watchedAt` with unsorted timestamps

PR #3079, opened 2026-08-04
fix(stats): scope monthly stats header and fix streak subtitles

tysonkerridge:
Current/previous streak are all-time totals, not month-scoped, so grouping them under "Monthly Stats" alongside dropped-streak and active-day counts was misleading.

- Moves the "Monthly Stats" header to sit only above the cards that are actually month-scoped
- Swaps "Keep it going!" onto Current Streak (only makes sense for an active streak) and gives Previous Streak its own "Ended" subtitle

Depends on https://github.com/trakt/trakt-web/pull/3078 — based on that branch, not `main`.

Before/After:
Screenshot 2026-08-04 at 3 52 15 pm Screenshot 2026-08-04 at 3 51 42 pm

cc @ElMagnea

### DIFF
--- a/projects/client/i18n/meta/en.json
+++ b/projects/client/i18n/meta/en.json
@@ -8460,7 +8460,11 @@
     },
     "text_stats_keep_it_going": {
       "default": "Keep it going!",
-      "description": "Motivational sub-label shown below the previous streak stat."
+      "description": "Motivational sub-label shown below the current streak stat."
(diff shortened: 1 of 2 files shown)


seferturan (approved):
I applied a small fixup to increase the distance between the cards and the legend, so it's more clear the legend belongs to the heatmap.

PR #3084, opened 2026-08-04
Media: use new translation language query param

rectifyer:
# Summary

Bumps `@trakt/api` to 0.6.0 and migrates the translations calls to the new contract shape, where the `language` filter is an optional query param instead of a path segment (trakt/trakt-api#901, refs trakt/trakt-api#869). The client now requests `/translations?language=xx` instead of `/translations/xx`.

# Changes

- `@trakt/api` pin bumped from 0.5.4 to 0.6.0 in `projects/client/package.json`.
- `movieIntlQuery`, `showIntlQuery`, and `episodeIntlQuery` pass `language` via `query` instead of `params`.
- MSW handlers: the per-language `/translations/{lang}` mocks (movie en/nl/pt, show en/nl/ja, episode en/nl) are collapsed into single `/translations` handlers that serve the same mock data keyed off the `language` query param, since the old path-based handlers would no longer match the new request URLs.

# Dependencies

Draft until the upstream pieces land, in order:

1. The Rails API deploys support for `?language=` on the translations endpoints.
2. trakt/trakt-api#901 merges and `@trakt/api` 0.6.0 is published (tag `v0.6.0`).
3. Then this PR: refresh `deno.lock` with an install (the 0.6.0 pin cannot resolve before publish) and mark ready for review.

Merging early would fail install in CI on the unresolvable 0.6.0 pin, and running ahead of the Rails deploy would return unfiltered translations because the query param is ignored.

### DIFF
--- a/projects/client/package.json
+++ b/projects/client/package.json
@@ -81,7 +81,7 @@
     "@tanstack/query-core": "5.101.4",
     "@tanstack/query-devtools": "5.101.4",
     "@tanstack/query-persist-client-core": "5.101.4",
-    "@trakt/api": "npm:@jsr/trakt__api@0.5.4",
+    "@trakt/api": "npm:@jsr/trakt__api@0.6.0",
     "@ts-rest/core": "3.52.1",
     "@use-gesture/vanilla": "10.3.1",
     "bits-ui": "2.18.1",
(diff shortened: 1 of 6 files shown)


seferturan:
Closing for now, we can tackle this when we do have an actual 0.6.0 and update the package. I think there have been some more changes that need to be dealt with anyway.

PR #3087, opened 2026-08-05
feat: Where to Watch drawer UI revamp

michaldrabik:
- horizontal scrolling in a narrow vertical space with only ~3 tiles visible is dreadful
- proposing a vertical based layout with foldable sections per streaming service
- now also matches UI style of how streaming services tiles are presented in the Settings
- bonus: no more grey scrollbars for browsers other than Chrome

Current:
image

After:
image

After Video:
https://github.com/user-attachments/assets/df784de4-b8b9-437b-9da6-3a56a37318c2

### DIFF
--- a/projects/client/i18n/meta/en.json
+++ b/projects/client/i18n/meta/en.json
@@ -720,6 +720,18 @@
       "default": "Search people",
       "description": "Aria-label for the people drawer search input."
     },
+    "input_placeholder_search_streaming_services": {
+      "default": "Search streaming service or country...",
+      "description": "Placeholder text for the where-to-watch drawer search input. This should be short and indicate users can search locally by country name or streaming service name."
+    },
+    "input_label_search_streaming_services": {
+      "default": "Search streaming services",
+      "description": "Aria-label for the where-to-watch drawer search input."
+    },
+    "button_label_toggle_streaming_countries": {
+      "default": "Toggle countries for {service}",
+      "description": "Aria-label for the button that expands or collapses the list of countries a streaming service is available in."
+    },
     "button_text_join_trakt": {
       "default": "Join Trakt",
       "description": "Text for the Trakt login button. This should be a short and concise message that indicates what users can do."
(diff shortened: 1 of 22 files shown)


michaldrabik:
Bonus idea - preview available top countries with flags strip.
This is not committed as I don't know if it's not to much -> need second opinions.

image

seferturan:
This looks great ❤️

michaldrabik:
- since the data is loaded as whole I also added local search just like we have for people in season details drawer (you can now quickly filter by country name or service)

Final proposal look below. Feel free to add or revert any changes during rev.

image

---

image

rectifyer:
@michaldrabik I like this change a lot as well. I did notice a small display bug, presumably if there is not image being returned. Notice for **Pilot WP** the text is left aligned since there is no image. I think it should still render the area and use text as the fallback (I thought it would do this already actually).
 
image

michaldrabik:
@rectifyer O yeah. So I think we could always display the tile logo background for the same spacing and now it's just question if to put the name in there or simply just some icon or Trakt icon as generic placeholder.

michaldrabik:
@rectifyer Added text fallback

seferturan:
Love the design, but as I was testing it out, some slight annoyances:

- If something is in available in your country, it will now always take an extra click to expand it first.
- Similarly, if it's only available in a single country, you need to expand it first.

Ideas (I left it as a separate second fixup commit atm):
- If it's available in your country, always make it available on the root. The expand can be used to view the other countries.
- If it's only available in one country, make it available in the root, no expand possible.
- Gave the one from your own country a subtle highlight to clearly differentiate it.

Examples:
Screenshot 2026-08-13 at 12 54 45

Screenshot 2026-08-13 at 12 54 54

@michaldrabik @ElMagnea @rectifyer thoughts?

michaldrabik:
Sure looks OK
Just 2 UI things in that case:

- can try making the logo image tile a little bit bigger now which will make the tile taller and give inner components a bit more space to breath between vertical edges and service title and country tile. At the same time logo will also be a bit bigger.

- items with flags only are now all highlighted and clickable to expand -> items with your country can only be expanded by clicking the "collapse" button. I would unify that to either one.

seferturan:
Updated version; less containers, clear separation of click zones:

image

image

seferturan (approved):
Applied some fixups:

- Changed the UX a bit to still allow for single clicks (e.g. not having to always expand a service).
- Got rid of some code duplication.

PR #3092, opened 2026-08-06
fix(stats): show full numbers below 100K instead of abbreviating

tysonkerridge:
## Summary

All-time profile stats (plays, episodes watched, ratings, comments, etc.) abbreviated large counts into compact notation like "1.2K", hiding the exact value users wanted to see. Since these counts realistically never exceed 100K, the fix shows the full grouped number below that threshold and only falls back to compact notation for the rare outlier above it, via a new `toHumanCount` helper that leaves the shared `toHumanNumber` (used elsewhere in the app for values that do routinely get large) untouched.

## Alternatives considered

- **Tap/hover tooltip on the abbreviated value**, matching the existing watch-streak heatmap interaction — prototyped (see attached screenshot) but dropped, since abbreviation essentially never triggers for these counts in practice, the tooltip plumbing would be unused complexity for a case that doesn't happen.
- **Lowering the abbreviation threshold globally in `toHumanNumber`** — rejected in favor of a separate `toHumanCount` helper, since `toHumanNumber` is used elsewhere in the app for values that do routinely get large, and changing it would be a much broader behavioral change than this fix calls for.

## Notable context

In several locales (Arabic, Hindi, Russian, Portuguese-BR) the abbreviated form is actually *wider* than the plain number in this range, since their thousand-suffix is a full word (e.g. "тыс.", "हज़ार") rather than a single letter — so abbreviating bought nothing there anyway, beyond the precision loss.

## Test plan

- [ ] Unit tests pass: `deno task test:unit` (covers `toHumanCount` across the 100K boundary and locale variation)
- [x] Visit a profile's all-time stats card and the all-time stats drawer; verify counts under 100K render as full grouped numbers (e.g. `4,217` not `4.2K`)
- [x] Sanity-check a non-`en` locale (e.g. `de` or `fr`) to confirm grouping separators render correctly
- [ ] Confirm the rare case above 100K still abbreviates (can temporarily lower `COMPACT_THRESHOLD` locally to exercise this, or check the new spec)

## Screenshots

Profile Small Before/After (en/ru)
en s before
en s after
ru s before
ru s after

Profile Large Before/After (en/ru)
en l before
en l after
ru l before
ru l after

Sheet Before/After (en/ru)
en sheet beforeen sheet after

ru sheet beforeru sheet after

Alternative - Tooltip (en):
tooltip

cc @ElMagnea @michaldrabik

### DIFF
--- a/.agents/rules/utils.md
+++ b/.agents/rules/utils.md
@@ -95,6 +95,7 @@ toRelativeHumanDay(today, date, localeKey);
 
 // number
 toHumanNumber(value, locale); // compact notation: 1.2k, 4.5M
+toHumanCount(value, locale); // counts: grouped below 100K (1,234), compact above
 toPercentage(value, locale); // 0.75 -> "75%"
 toTraktRating(rating, locale); // alias for toPercentage
 toHumanCurrency({ price, currency, locale });
--- a/projects/client/src/lib/sections/profile/components/_internal/AllTimeStatTile.svelte
+++ b/projects/client/src/lib/sections/profile/components/_internal/AllTimeStatTile.svelte
@@ -2,7 +2,7 @@
   import LockIcon from "$lib/components/icons/LockIcon.svelte";
   import KpiTile from "$lib/components/kpi/KpiTile.svelte";
   import { languageTag } from "$lib/features/i18n";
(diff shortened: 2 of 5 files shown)


seferturan (approved):
Applied some minor fixups to remove some comments and get rid of a `$derived`

PR #3099, opened 2026-08-07
fix(i18n): stop baking a literal % into season-ratings message templates

tysonkerridge:
## Summary

`text_ratings_season_extremes` and `text_ratings_season_point` (the summary line and tooltip on the season-ratings chart in the ratings breakdown drawer) hardcoded a literal `%` after the rating placeholder, which assumes English's no-space convention. Locales that require a space before the percent sign (German, French, ...) were rendering it incorrectly in every translation. The rating is now formatted by the caller via `toPercentage()` before being interpolated into the message, so the sign's locale-specific spacing comes from `Intl.NumberFormat`, not a literal baked into the template. This matches a fix already shipped on iOS for the same message keys.

Also corrects `text_ratings_season_extremes_android`'s description, which incorrectly documented `{peak}`/`{low}` as pre-formatted labels like `"S4"` when the template itself prepends `S` (they're raw season numbers) — no functional change, since `description` isn't read by the generator, just fixing stale docs while touching this key's neighbor.

## To check

- [ ] @seferturan will the translations get updated as expected to remove the raw % from all of them locally/at crowdin?
- [ ] @michaldrabik was there a reason android uses a different string instead of updating the existing?

## Screenshots
Before/After (German, which has spaces between `number` and `%`)
Screenshot 2026-08-07 at 3 42 36 pm
Screenshot 2026-08-07 at 3 42 05 pm

cc @ElMagnea @michaldrabik

Ref: [Android](https://github.com/trakt/trakt-android/pull/332)

### DIFF
--- a/projects/client/i18n/meta/en.json
+++ b/projects/client/i18n/meta/en.json
@@ -2636,8 +2636,8 @@
       "description": "Accessible label for the season-ratings line chart in the ratings breakdown drawer."
     },
     "text_ratings_season_extremes": {
-      "default": "Peak {peak} · {peakRating}% · Low {low} · {lowRating}%",
-      "description": "Summary line above the season-ratings chart calling out the highest- and lowest-rated seasons. {peak}/{low} are compact season labels (e.g. S4); {peakRating}/{lowRating} are whole-number percentages.",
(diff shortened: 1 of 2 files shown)


seferturan (approved):
@tysonkerridge, if not I'll do a manual trigger for crowdin 👌

PR #3100, opened 2026-08-07
feat: independent comment language filter

michaldrabik:
The comment/review language filter will no longer derive its default from the app
locale. It now defaults to **All languages** and remembers the user's choice
across sessions and navigation.

### Why?

- Defaulting the filter to the app language meant users browsing in a non-English
locale silently saw a heavily filtered review list - usually always empty (we don't have that big of volume to cover our languages) - with no indication that a filter was applied. Reviews are mostly written in English,
so the app-language default hid the majority of content.

- Language preference for reviews is also a separate concern from UI
language: someone using a Polish interface may still want to read reviews in whatever language has most of them in general. Tying the two together conflated two unrelated settings.

- Finally, the filter was URL-only, so it reset on every navigation. Picking a
language on one media had to be repeated on the next one. Persisting it matches
how other user-level view preferences (collapsed sections, sort togglers)
already behave.


seferturan (approved):
Applied a fixup to still default to the locale instead of 'all' languages

PR #3101, opened 2026-08-07
fix(tags): keep the poster tag at its original size and clear the info row

ElMagnea:
Follow-up to #3053, which shipped two size changes where only one was intended.

## The tag itself was being scaled, not just the glyph

`IndicatorTag` carried `scale: 1.15` on `.trakt-tag`, which scales the whole pill — background, padding, radius and glyph together — taking it from 22×18px to 25.3×20.7px. That is separate from the per-glyph `--glyph-scale`, which is the mechanism that actually needed to change: the double-check fills only ~80% of its viewBox where the bookmark fills 100%, so it rendered short next to it.

Only the glyph correction was wanted. Dropping the pill scale leaves every `--glyph-scale` untouched, so the checkmark keeps its corrected height, and the pill returns to the size it has always been. The tag radius also lands at exactly the 6px the summary poster pills use, rather than 6.9px.

## Footer clearance is now derived instead of hand-tuned

`IndicatorTags` centres the tag on the cover's bottom edge, so half of it hangs into the footer underneath. The padding compensating for that was a literal in `CardFooter`, and the room for it a second literal in `--height-card-footer-sm`. Nothing tied those to the tag's actual height, so any resize left them out of step and the info row crept back up against the tag — on small cards it was down to ~2px of gap.

`--indicator-tag-overhang` is now computed from the tag's own box, and `CardFooter` pads by that plus a gap:

| footer | before | after |
| --- | --- | --- |
| normal (40px) | 8.0px gap | 10.5px gap |
| small (28px → 36px) | 2.0px gap | 8.5px gap |

Small cards grow 8px as a result, and every poster list height derives from that token, so Related / Recommended / Smart lists grow with it.

## Testing

`deno fmt --check` clean · `svelte-check` 0 errors / 0 warnings across 9358 files. The identical CSS changes were also run against the full unit suite (2684 passing) on a branch sharing this file state — the suite could not run in the worktree used here, as Vite's `/@fs` resolver cannot follow its symlinked `node_modules`.

### DIFF
--- a/projects/client/src/lib/components/media/tags/IndicatorTag.svelte
+++ b/projects/client/src/lib/components/media/tags/IndicatorTag.svelte
@@ -58,8 +58,6 @@
       height: var(--ni-10);
       line-height: var(--ni-10);
 
-      scale: 1.15;
-
       :global(trakt-tag-icon svg) {
         width: var(--ni-10);
         height: var(--ni-10);


seferturan (approved):
For now I reverted the changes on the footer, it would increase the spacing too much on pretty much all lists. Example:

Image

PR #3103, opened 2026-08-07
fix(spoiler): keep the spoiler box open when expanding the description

### LINKED ISSUE #3081: Can't view whole description in a spoiler alert box
### Code of Conduct

https://app.trakt.tv/shows/the-graham-norton-show?season=33&view=episode&episode=6

Clicking on "more" button inside a spoiler alert box collapses the spoiler alert box instead of expanding the description. See video below.

Tested on Chrome 150 on Android.

### Additional information

https://github.com/user-attachments/assets/b65a9b41-1b04-4cb8-9d07-3673d2ae3c60

rectifyer:
# Summary

Clicking the "more" button inside an expanded spoiler alert box collapsed the whole box instead of expanding the clamped description. The `contain` variant of `CollapsableContent` rendered its content inside the toggle `<button>`, so the nested "more" button was invalid HTML (interactive content inside a button) and its click bubbled up to the container's toggle.

Fixes #3081

# Changes

`CollapsableContent` now renders the `contain` variant as a wrapper `<div>` that carries the spoiler box styling (border, background, color transitions), with the header as the only `<button>` and the content as its sibling. Clicks inside the description no longer reach the toggle, and "Hide description" in the header still collapses the box.

Visual parity is preserved: the header keeps its full-width space-between layout via `flex: 1`, and the content padding compensates for the button padding it previously inherited so the combined inset is unchanged. The toggle button also gains `aria-expanded` since expansion state is no longer conveyed by nesting.

`EpisodeInfoHeader` stretches the spoiler box to full width via a global selector, so the new container class was added alongside the existing button selector there.

# Testing

Verified in the running app on an episode with a long spoiler-hidden description: expanding the description with "more" keeps the box open, header alignment matches the previous layout, and collapsing via the header still works.

### DIFF
--- a/projects/client/src/lib/sections/summary/components/_internal/CollapsableContent.svelte
+++ b/projects/client/src/lib/sections/summary/components/_internal/CollapsableContent.svelte
@@ -6,13 +6,11 @@
     children,
     labels,
     headerContent,
-    variant = "sibling",
     isCollapsed,
     toggle,
   }: {
     labels: { view: string; hide: string };
     headerContent?: Snippet;
-    variant?: "contain" | "sibling";
     isCollapsed: boolean;
     toggle: () => void;
   } & ChildrenProps = $props();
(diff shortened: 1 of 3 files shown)


seferturan (approved):
pplied some fixups:

  - Dropped the `sibling` variant. Never had a consumer, `SpoilerSection` always passed `contain`.
  - Inlined both snippets, single use now. Flattened the css along with it.
  - Dropped the redundant `:global(.trakt-collapsable-content-button)` width rule in `EpisodeInfoHeader`, the container stretches it anyway.
  - Dropped the `aria-label`. It duplicated the visible label and hid the spoiler alert text from the accessible name.

PR #3113, opened 2026-08-10
fix(settings): handle per-user plex server errors inline instead of full-page

rudf0rd:
## What

When `GET /users/settings/plex/servers/:id` fails with a per-user error — most commonly 503 `invalid_server_url` (Trakt can't reach the user's Plex media server) — the Plex settings page no longer gets replaced by the global **“Trakt service unreachable”** page. The failure now renders inline on the server card with actionable copy and a Retry button.

## Why

`plexServerAccountsQuery` goes through `defineQuery`, so any non-200 throws a `FetchError` and dispatches the global fetch-error event. `mapToWellKnownError` maps every 503 to `ServerError`, which swaps the entire page for the outage screen. Since every settings write invalidates `Plex.Settings` (which this query subscribes to), each toggle re-fired the 503 and re-took-over the page — making Plex settings unusable for any user whose server has remote access disabled or port 32400 blocked. Reproduced in prod with an affected user (server reports 4 connections, all local-only → `url: null` → Rails 503 `invalid_server_url`).

This is per-user Plex state, not a Trakt outage, and the API already says so in the body (`error_code`, `message`, `guidance`).

## How

- `isValidResponse`/`FetchError`: the failing query's key now rides the fetch-error event as `source` (e.g. `query:plexServerAccounts`), through `CustomFetchError` → `mapToWellKnownError` → `WellKnownError`.
- `ErrorProvider`: exempt errors are dropped at capture instead of filtered at render, so an exempt failure can never overwrite (and dismiss) a non-exempt error already showing the global page.
- `queryId.ts`: uplifted from `query/_internal/` to the query feature's public surface, since the exemptions now reference it cross-feature.
- `errorExemptions.ts`: exemptions gain an optional `sources` filter. `/settings/plex` exempts **only** `plexServerAccounts` failures from the global `ServerError`/`NotFoundError` pages — a failing `plexSettingsQuery` (or anything else on the route) still gets the global page, so no component is left in a permanent loading state. The pre-existing streaming-services exemptions stay route-wide.
- `usePlexServer`: expose `isAccountsError`, `accountsErrorCode`, and `retryAccounts`.
- `toPlexErrorCode.ts`: extracted body `error_code` digging into a testable helper.
- `PlexServerCard`: inline `role=alert` error state — dedicated copy for `invalid_server_url` (“enable remote access / open port 32400”), generic load-failed copy otherwise, plus Retry.
- New i18n key `error_text_plex_server_unreachable` in `i18n/meta/en.json` (translations land via the Crowdin localize workflow post-merge).

## Testing

- New: `errorExemptions.spec.ts` (route/source/type matching incl. the scoped plex entries), `toPlexErrorCode.spec.ts` (extraction + malformed shapes)
- `deno task check` (svelte-check): 0 errors; eslint clean on touched files
- vitest `src/lib/requests/plex` + `src/lib/features/errors`: 22 passed

Pairs with trakt-rails#2395 (plex-side auth failures 401 → 403), which fixed the logout loop this page previously caused.


rudf0rd:
after reviewing again, this solution is too heavy handed for what i need.

PR #3114, opened 2026-08-10
fix(settings): treat per-user plex server failures as data, not errors

rudf0rd:
---

<div align="center">

### Plex server errors stay on the card, not the whole page

`fix` · _An unreachable Plex media server now reads as data about that server — rendered inline with retry — instead of a full-page Trakt outage._

</div>

---

## 🎬 Scene Setting

When a user's Plex media server is unreachable (remote access off, stale server id, plex timeout), the API answers the server-accounts endpoint with a 404/5xx carrying the structured `{error_code, message, guidance}` envelope. Our global error handler read that status as a Trakt outage and replaced the entire settings page — for something that is really just information about the user's own server.

This PR moves the classification to the request boundary instead of teaching the error system to forgive:

- `plexServerAccountsQuery.ts` parses the body against the SDK's `plexErrorResponseSchema`; an envelope match is normalized into the query result as an `{ errorCode }` data variant. A bare error response without the envelope (a genuine outage) still flows through as a query error and raises the global page.
- `PlexServerCard` renders the failure inline: bold red headline, secondary guidance line, and a Retry button (with spinner while the refetch is in flight).
- `SyncLoadError` is uplifted from `streaming-services/` to `settings/_internal/` and shared by both settings surfaces, with new `message`/`hint`/`variant` props and the retry button's aria-label corrected to `button_label_retry`.

The global error pipeline is untouched — no exemptions, no route/query coupling.

## 📸 Show, Don't Tell
CleanShot 2026-08-10 at 15 56 18@2x

## 🧪 Testing

- New spec cases on `plexServerAccountsQuery`: a 503 with the plex envelope resolves to `{ errorCode }` data; a 503 without it stays a query error.
- Full unit suite green (2750 tests) and `svelte-check` clean.
- Verified manually against the dev server with a mocked 503 envelope: inline card error with working retry, and the bare-503 path still shows the global error page.

Refs #3113


seferturan (approved):
Added some fixups: 

- Narrowed `errorCode` to the union the backend actually returns, with an `unknown` fallback for new ones.
- Added copy for the remaining codes, we only handled `invalid_server_url` before.
- Hidden the retry button where retrying can't help (bad auth, no server selected, missing server).
- `usePlexServer` returns a single `accountsError` with `retry` on it, instead of three separate values.
- `isLoadingAccounts` goes through `toLoadingState` now.
- Extracted `refetchQuery`, both streaming services sections use it too.

PR #3124, opened 2026-08-12
fix(navigation): move the signed-in dashboard to /home

seferturan:
## 🎶 Notes 🎶

- The landing page could briefly flash for signed in users on `/`. Both the landing and the dashboard lived in the same route, so whenever auth resolved late the visible content repainted in place.
- `/` is now landing only, `/home` holds the dashboard. Server loads redirect both ways off the oidc cookie, so the bounce lands before first paint instead of after.
- Server and client now agree on what an authorized session is (token + expiry, one shared helper). The cookie outlives the oidc storage the client reads (ITP evicts it after 7 idle days), and presence only checks had the server answering "authorized" while the client rendered as public, so the two bounced off each other forever.
- The callback awaits the token post before leaving, otherwise `/home` turns the request away and we land back on the flash we're removing.
- Service worker: `/` is network first, and redirected documents are never cached. A signed out `/home` fetch follows the 307 and comes back as the landing document, which would replay the same flash one route over.
- Redirect responses are stamped `no-store`. Cloudflare keys on URL and doesn't vary on cookie, so a cached `307 / -> /home` would bounce anonymous visitors back and forth and leave the landing page unreachable.
  - This applies to every load level redirect, not just the audience ones. Keeping it broad on purpose: an allowlist of guarded paths fails open the moment a route gets renamed.
- `UrlBuilder.home()` now means `/home`, `landing()` is the new name for `/`. PWA `start_url` follows the dashboard since installs are pinned by manifest id.
- Query params are preserved across the server guards, same as the other `load` redirects. The client side fallback in `TraktPage` drops them.
- ⚠️ The flash was never reliably reproducible, so the fix is reasoned from the mechanism rather than observed. We should keep an eye on prod after this deploys.

### DIFF
--- a/projects/client/src/hooks.server.ts
+++ b/projects/client/src/hooks.server.ts
@@ -11,6 +11,7 @@ import { handle as handleLegacyRedirect } from '$lib/features/legacy-redirects/h
 import { handle as handleMobileOperatingSystem } from '$lib/features/mobile-os/handle.ts';
 import { handle as handleSearchConfig } from '$lib/features/search/handle.ts';
 import { handle as handleTheme } from '$lib/features/theme/handle.ts';
+import { isAuthorizedToken } from '$lib/features/auth/isAuthorizedToken.ts';
 import { isBotAgent } from '$lib/utils/devices/isBotAgent.ts';
 
 import { SENTRY_DSN } from '$lib/utils/constants.ts';
@@ -29,6 +30,10 @@ const WHITELISTED_HEADERS = new Set([
   'x-pagination-page',
   'x-pagination-page-count',
 ]);
+
+const NO_STORE = 'private, no-store, no-cache, must-revalidate';
+
+const REDIRECT_STATUSES = new Set([301, 302, 303, 307, 308]);
 
 function hasWebviewParam(url: URL): boolean {
   return Object.values(WEBVIEW_PARAMS).some((param) =>
@@ -55,16 +60,24 @@ export const handleReferrerPolicy: Handle = async ({ event, resolve }) => {
 export const handleCacheControl: Handle = async ({ event, resolve }) => {
   const response = await resolve(event);
 
-  if (!response.headers.get('content-type')?.includes('text/html')) {
+  const isHtml = response.headers.get('content-type')?.includes('text/html') ??
+    false;
+  const isRedirect = REDIRECT_STATUSES.has(response.status);
+
+  if (!isHtml && !isRedirect) {
     return response;
   }
 
   function toCacheControl() {
+    if (isRedirect) {
+      return NO_STORE;
+    }
+
     // A WebView request carries the viewer's VIP token in the URL. Never let a
     // spoofed bot User-Agent promote the response to a public CDN entry (keyed
     // by the token), which would cache one viewer's review for others.
     if (hasWebviewParam(event.url)) {
-      return 'private, no-store, no-cache, must-revalidate';
+      return NO_STORE;
     }
 
     // Verified search engine crawlers (via Cloudflare), full CDN caching for SEO
@@ -76,13 +89,13 @@ export const handleCacheControl: Handle = async ({ event, resolve }) => {
     // Only cache publicly for unauthenticated requests to prevent cache poisoning via spoofed User-Agent
     if (
       isBotAgent(event.request.headers.get('user-agent')) &&
-      !event.locals.oidcAuth
+      !isAuthorizedToken(event.locals.oidcAuth)
     ) {
       // 120 seconds is enough to satisfy Discord without heavily caching stale content
       return 'public, max-age=120, s-maxage=120';
     }
 
-    return 'private, no-store, no-cache, must-revalidate';
+    return NO_STORE;
   }
 
   const clonedHeaders = new Headers(response.headers);
--- a/projects/client/src/lib/components/router/Redirect.svelte
+++ b/projects/client/src/lib/components/router/Redirect.svelte
@@ -5,7 +5,8 @@
   const { to }: { to: string } = $props();
 
   onMount(() => {
-    if (page.url.pathname === to) return;
+    const target = new URL(to, page.url.origin);
+    if (page.url.pathname === target.pathname) return;
     goto(to, {
       replaceState: true,
     });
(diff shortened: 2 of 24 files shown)


rectifyer:
# Review notes

The design holds up overall. I verified the redirect/no-store interplay (legacy 301s keep their CDN cacheability because `handleLegacyRedirect` short-circuits before `handleCacheControl`), the service worker caching (`skipRedirectedDocuments` correctly keeps the followed-307 landing document out of the `/home` cache key, and sharing one `ExpirationPlugin` between the two navigation strategies is safe since both use the same cache name), and the manifest `id` pinning that makes the `start_url` change safe for existing installs. `TraktPage` with `audience="public"` is only used by the landing route, so the new redirect map doesn't lock signed-in users out of any other page.

That said, one regression and one failure-mode risk stand out, plus a few smaller items.

# Large

## Every signed-out page load now POSTs to `/api/store-token`

In `initializeUserManager.ts`, the boot path for a signed-out visitor now calls `handleUserEvent(null)`, and `handleUserEvent` calls `postToken` unconditionally. The old code went through `setAuthState` + `syncToken`, which early-returns on a null user, so anonymous boots never posted.

Every anonymous visitor now fires a POST (with up to 3 retries) plus a `Set-Cookie` on every full page load, even when there is no cookie to clear. The desync being fixed only needs the POST when the server thought the session was authorized, so gating on `tokenFromServer` keeps the fix and drops the wasted traffic:

```ts
if (!user) {
  if (tokenFromServer) {
    handleUserEvent(null); // clear the stale cookie
    return;
  }
  setAuthState({ token: mapToToken(null), isExpired: true });
  return;
}
```

The event-listener paths (`userUnloaded` / `silentRenewError`) should keep posting unconditionally, as they do now.

## Unbounded redirect ping-pong when `/api/store-token` fails

Convergence rests entirely on `postToken` succeeding. If it exhausts its 3 retries, `.catch(() => null)` swallows the failure and nothing ever retries. At that point the two guards fight each other with no loop breaker:

- Cookie says authorized, client signed out (the ITP case): `/home` renders, `TraktPage`'s `<Redirect>` sends the client to `/`, the server load sees the still-valid cookie and 307s back to `/home`. `Redirect.svelte`'s `pathname === to` guard never trips because the two paths differ, so it loops indefinitely.
- The mirror case (client authorized, cookie expired or missing) loops the same way in reverse until `syncToken`'s POST lands.

Normally the POST wins within a bounce or two, so this is an edge case, but the failure mode is an endless navigation loop rather than a stale page. A small breaker would cap the damage, e.g. a sessionStorage bounce counter in `Redirect`/`TraktPage`, or only firing the client-side audience redirect once per session.

# Small

- **Query params are dropped on the client-side guard.** `redirectForAudience` appends `url.search`, but `TraktPage`'s `<Redirect to={UrlBuilder.home()}>` does not. A signed-in user landing on `/?utm_...` client-side loses the params the server path would have kept, despite the PR notes saying both guards preserve them.
- **Double POST on login.** `signinCallback()` fires `userLoaded`, whose listener already posts the token via `handleUserEvent`; then `storeSession` in `callback/+page.svelte` posts it again. Harmless (idempotent cookie set), and the awaited one guarantees ordering, but it is two requests per sign-in.
- **`/home` lost `type="home"` and `info`.** The new `home/+page.svelte` omits them, so `type` defaults to `"webpage"`: the og:title format changes and the meta description falls back to the default overview. The page is `noindex` so this is mostly cosmetic, and keeping the WebSite JSON-LD on the public landing is arguably better for SEO. Just confirming it was deliberate.
- **`NetworkFirst` for `/` has no `networkTimeoutSeconds`.** On a flaky connection the landing now hangs on the network instead of serving the cached copy (StaleWhileRevalidate used to answer instantly). A few-second timeout would keep the freshness goal without the worst-case stall.
- **Inconsistent auth check in `hooks.server.ts`.** The social-bot cache branch still gates on cookie presence (`!event.locals.oidcAuth`) while everything else now uses `isAuthorizedToken`, so an expired cookie makes a bot request uncacheable. It fails closed, so it is safe, just at odds with the new shared-helper story.

seferturan:
**@rectifyer from your review:**

  - `initializeUserManager` only posts the empty token when the server actually sent one. Anonymous boots
  were firing a POST plus a `Set-Cookie` on every full page load, straight regression.
  - `postToken` shares its in-flight request, so the sign-in callback awaits the listener's POST instead of firing a second one. Fixes the double POST and saves a roundtrip on login.
  - Added `networkTimeoutSeconds` on the landing page (1.5s), so a flaky connection falls back to the cached landing instead of hanging.
  - Bot cache branch uses `isAuthorizedToken` now, so it lines up with everything else. Side effect: an expired cookie plus a bot UA gets the 120s public cache again, which is correct since the server renders that request as anonymous anyway.
  - Client guard carries the query string on `public -> home`. Left it off `authenticated -> landing`, that one is a rejection so dragging `/calendar?date=...` onto the landing page is meaningless.
  - Fixed the PR notes, they claimed both guards preserve query params. Only the server ones do.
  - `/home` dropping `type="home"` and `info` was deliberate, keeps the WebSite JSON-LD on the public landing only.

  **Other things changed:**

  - **The ping-pong is fixed structurally.** The server guard now only runs on document requests (`isDataRequest`), so the two guards act on disjoint navigations and can't contest each other. The client owns client side navs, which is right since oidc-client-ts is the actual session and the cookie is just a derived cache for server side routing.
    - Desync case is now one clean navigation instead of a bounce, and no counter, no sessionStorage, no threshold. Went down the counter route first and it doesn't hold up: it blocks legit repeated redirects (signed out user clicking calendar -> profile -> history gets stuck on the third).
    - Only real tradeoff: a signed in user client navigating to `/` gets an empty shell for a frame before the client guard moves them. Nothing in the authed UI links to `/`, so this needs the back button or a hand typed URL.
    - Verified with curl: document requests still 307 both ways, data requests skip.
  - Added a spec for `redirectForAudience`. It's the load bearing guard and had no coverage.
  - `Redirect` compares pathnames now that `to` can carry a query. Worth noting `shows/:slug/seasons/:season` already passes one today, checked both non bare call sites and they target a different pathname, so behaviour is unchanged.

rectifyer (approved):
Looks good. Tested locally with sign in, sign out, navigating around, private browser, etc and working as expected.

PR #3126, opened 2026-08-13
fix(settings): prevent raw export from hanging on stalled requests

rudf0rd:

A member reported the raw export hanging forever on the collection step - no error, still spinning after running overnight. Their account turned out to be a mega collection: **~484k collected items** (~56k movies, 14.5k shows, ~428k episodes), which means the export's collection section alone is **~2,000 sequential paginated requests**.

Three problems compounded:

1. **No timeout anywhere in the export fetch chain.** Browser `fetch` has no default response timeout, so one stalled response stream out of those 2,000 leaves `await response.json()` waiting forever. Nothing throws, so the retry loop never fires and no error ever surfaces. This is the silent hang.
2. **Every page was retained pretty-printed** (`JSON.stringify(data, null, 2)`) as a `Uint8Array` until the final in-memory zip. Measured on the reporter's real pages: 1.5-2x inflation, ~514MB retained for the collection section.
3. **No page-level progress.** The status text sits on the same endpoint name for 30+ minutes on a big account, indistinguishable from a hang.

The fix, in order: each fetch attempt now aborts via `AbortSignal.timeout(1min)` and feeds the existing ts-retry loop (10 tries, 10s delay) - a stall now ends in a visible "Export failed" instead of an infinite spin; pages are stored compact (**514MB → 332MB retained, -35%**, measured on the same account); and paginated endpoints show the current page number in the progress text. Page *totals* are deliberately not shown - the collection page-count header reports the combined collection size, not the per-type count.

Not addressed here (workers-side, will file separately): intermittent 500s on `collection/shows` for mega collectors, and the combined item-count pagination headers.

Text-only UI change - progress now reads `(44/50 · 412)` (endpoint counter · current page) instead of `(44/50)`.

- **`fetchWithRetry.spec.ts`** (new): happy path parses payload + pagination headers; a stalled response (MSW `delay('infinite')`) rejects instead of hanging; a 500 retries and recovers. Retry knobs are injectable with production defaults so the failure paths run in milliseconds.
- **Negative control:** with the timeout neutered, the stalled-response test hangs until vitest kills it at 5s - the spec genuinely catches the regression.
- **Full-scale replay against production:** re-ran the collection section for the affected account (public data, read-only) through this exact logic - pagination termination, retry policy, 60s timeout, compact stringify, fflate zip. Results: **1,998 requests, 34.8min wall, 0 retries needed (0 timeouts / 0 429s / 0 5xx this run), 332MB fetched, 1,998 files zipped to 34MB, and the pagination loop terminated exactly where the data ends.** The 30+ min fetch phase at zero failures is also why the page-progress change matters - that's the window that previously looked like a hang.
- Full export dir suite: 11/11 passing; `deno fmt` clean.

### DIFF
--- a/projects/client/i18n/meta/en.json
+++ b/projects/client/i18n/meta/en.json
@@ -5034,6 +5034,27 @@
       "default": "Export complete.",
       "description": "Text shown to indicate that the user's data export is complete."
     },
+    "text_export_status_partial_one": {
+      "default": "Export complete, but 1 of {total} sections failed. See _errors.json in the zip.",
+      "description": "Text shown when the user's data export finished but exactly one section could not be fetched.",
+      "variables": {
+        "total": {
+          "type": "number"
+        }
+      }
+    },
+    "text_export_status_partial_other": {
+      "default": "Export complete, but {failed} of {total} sections failed. See _errors.json in the zip.",
+      "description": "Text shown when the user's data export finished but multiple sections could not be fetched.",
+      "variables": {
+        "failed": {
+          "type": "number"
+        },
+        "total": {
+          "type": "number"
+        }
+      }
+    },
     "header_import": {
       "default": "Import",
       "description": "Header for the settings section that allows users to import their data from third-party services."
--- a/projects/client/src/lib/features/analytics/events/AnalyticsEventDataMap.ts
+++ b/projects/client/src/lib/features/analytics/events/AnalyticsEventDataMap.ts
@@ -67,7 +67,11 @@ type ImportCompletedType = {
   duration: number;
 };
 type ImportFailedType = { source: string; error: string };
-type ExportCompletedType = { duration: number; endpointCount: number };
+type ExportCompletedType = {
+  duration: number;
+  endpointCount: number;
+  failedCount: number;
+};
 type ExportFailedType = { error: string };
 
 type ClearInitiatedType = { source: string };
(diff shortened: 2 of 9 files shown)


seferturan:
Applied some fixups:

- Partial exports are visible now. A failed section used to report `Export complete.` with no hint anything was missing. Status reads `Export complete, but 3 of 112 sections failed. See _errors.json in the zip.`, sticks around instead of auto-clearing, and reports `failedCount` on `ExportCompleted`.
- One failing endpoint no longer bins the whole export. Endpoints fail individually and `_errors.json` lists what's missing, plus how far each one got. Pages fetched before a failure are already in the zip, so "failed" on its own didn't say whether you lost one page or all 1,712.
- Covers the dynamic list discovery too, which used to kill the export before any endpoint ran.
- Pagination termination. Requested page is authoritative, the echoed `X-Pagination-Page` is gone entirely. A stale header would have looped forever, and every page would have overwritten page 1's file.
- Defensive rather than observed, the prod replay terminated exactly where the data ends.
- A garbage page-count header parsed to `NaN` and walked to the 10k cap issuing real requests. Falls back to a single page.
- 2h budget on the fetch phase. The retry budget is per page, so it multiplies: ~12 min worst case for one page, and collection/episodes wants 1,712 of them. Past the budget the rest fail fast and you still get everything fetched so far.
- Stops retrying 403/404. Was burning ~90s per endpoint to land on the answer the first response already gave.
- Pagination loops instead of recursing. One frame per page meant ~1,712 suspended frames each holding their own parsed page, which is the opposite of the point of this PR.
- `fetchWithRetry` takes a single object param, logs via `print.ts`.
- Status labels reset when you start another export.

seferturan (approved):

PR #3131, opened 2026-08-14
feat(seasons): overflow card + episode counts in the season dropdown

ElMagnea:
## What

Two fixes for the summary page's seasons strip silently hiding episodes:

### 1. Overflow card in the episode strip
The desktop strip is masked (not scrollable) and fits exactly N landscape slots — any episode past that was invisible with no hint it existed. Now, when a season has more episodes than the row fits, the last slot becomes a dashed **"+N more episodes"** card showing the hidden range (E7 – E10) and their combined runtime (3h 1m left). It links to the seasons drawer, same as the header drilldown.

- New `useLandscapeListItemCount` hook mirrors the `dynamic-item-count` SCSS formula (via `useVarToPixels`, live on resize), so the visible count always matches what the CSS row actually fits.
- Applies on tablet-lg/desktop only; mobile & tablet-sm keep the uncapped, free-scrolling strip.
- `SectionList` gains a generic `trailingItem` snippet rendered after the items.

### 2. Episode counts in the season dropdown
Season options now carry their episode count so you know what's behind each entry:
- `S1 - 10 Eps` (S prefix so it doesn't read as an episode range)
- `S1: Murder House - 9 Eps` for titled seasons
- `Specials - 4 Eps`

The orphaned `text_season_number_with_title` key is pruned from meta + all locale files.

## Screenshots

Strip with overflow card (Silo S1, 10 eps, viewport fits 7 slots) — dropdown open shows the new labels.

## Verification

- `deno fmt` + `svelte-check` (0 errors) + full vitest suite (2803 passed) + `i18n:check`
- Verified in the running app: strip renders 6 eps + ghost card, dropdown labels correct, card href matches the seasons-drawer drilldown


seferturan (dismissed):
This contained a lot of changes from your prototype branch and wasn't targeting main 😅 I took only the relevant commit and applied some fixups:

- I changed it to show the episode count using the existing translation keys.
- I put it between parentheses instead:
  - It's more informational, and not really related to the season number or title.
  - That we we can also reduce the delta, since we can keep everything else as is.
- I reverted the changes of the `more episodes card` for now. If we want that, we should tackle it separately and in a more generic way.

Example:

Image

seferturan (approved):

PR #3133, opened 2026-08-14
feat(settings): reset your cover image back to the default artwork

ElMagnea:
##Imagery
Screenshot 2026-08-14 at 16 39 22
Screenshot 2026-08-14 at 16 39 31

## The Setup

Setting a cover image is a one-way door. You can pick a new one from any movie,
show or episode page, but there is no way back to the default artwork — once a
cover is set, the only move is to set a different one.

This adds the way back: a **Cover Image** row in Settings → General →
Appearance, with a Reset that clears the custom cover and restores the default.

## How It Works

- The row only renders for signed-in users who actually have a cover set
  (`$user.cover.url`), so nobody gets a reset button for artwork they never
  customised.
- Reset asks first, through the existing confirmation machinery
  (`ConfirmationType.ResetCoverImage`, `destructive` operation) — the same
  flow as revoke-app and delete-list.
- On confirm it invalidates `InvalidateAction.User.CoverImage`, so the page
  background swaps back in place without a reload.

### About the endpoint

There is no delete route for covers. Probing the API returns `400` for `DELETE`
on both `/users/set_cover` and a nonsense path, while `PUT` answers `401` — so
`DELETE` simply is not routed. Per @seferturan, clearing a cover is the setter
itself with a sentinel id:

```
PUT /users/set_cover  { "cover_type": "show", "cover_id": 0 }
```

`resetCoverImageRequest` goes through the SDK's own `users.cover()` rather than
`rawApiFetch`, so there is no hand-written path to drift.

## One Extra Thing: `SettingsRowControl`

The three Appearance controls each sized to their own label, so Theme, Language
and Reset stacked at three different widths — and buttons default to `--ni-52`
against the select triggers' `--ni-40`, leaving Reset visibly taller than its
neighbours.

New `SettingsRowControl` gives them one shared track (width via
`--width-settings-row-control`, default `--ni-148`) and pins the button to the
trigger's height and inset. Long locale names ellipsize in the track; the full
name is still visible in the dropdown.

## Testing

- `resetCoverImageRequest.spec.ts` — asserts the exact request (`PUT
  /users/set_cover`, body `{ cover_type: "show", cover_id: 0 }`), so a silent
  drift in path or payload fails the suite.
- `ResetCoverImageRow.spec.ts` — clicks Reset, asserts the confirmation appears
  and that **no** request fires before you answer.
- This needed `ConfirmationProvider` in the component test bed. It lives in
  `ComponentTestBed` rather than `TestProvider`: `vitest-setup` drags
  `TestProvider` into every spec, and the dialog's import graph breaks specs
  that mock `$app/state` with a minimal `page` shape (it took 14 unrelated
  tests down before being scoped). Worth knowing — nothing in the suite had
  ever rendered a confirmation dialog until now.

`svelte-check` clean across 7314 files; full suite 2800 passed.

## Screenshots

_Added by @ElMagnea below._

---

## Pull Request Checklist

- [x] **Clear and Concise Title:** Does your title shine a spotlight on the
      essence of your changes?
- [x] **Detailed Description:** Have you set the stage with a compelling
      narrative that explains the "why" and "how" of your changes?
- [x] **Screenshots/Videos:** Have you captured the magic with visual evidence
      that showcases your changes in action?
- [x] **Tests:** Has your code undergone a rigorous dress rehearsal with unit
      tests to ensure it performs flawlessly under pressure?
- [x] **Coding Standards:** Does your code adhere to the project's style guide,
      ensuring it's as elegant as a film score?
- [x] **Commit Messages:** Are your commit messages clear and concise, guiding
      the audience through the evolution of your code?
- [x] **Documentation:** If your changes impact the script, have you updated the
      documentation accordingly?
- [x] **Code Review:** Are you prepared to embrace feedback from the discerning
      eyes of the project maintainers and fellow contributors?

### DIFF
--- a/projects/client/i18n/meta/en.json
+++ b/projects/client/i18n/meta/en.json
@@ -4403,6 +4403,34 @@
       "default": "Language",
       "description": "Text for the section that allows users to change the language."
     },
+    "text_settings_cover_image": {
+      "default": "Cover Image",
+      "description": "Title of the settings row that lets users reset the custom cover image shown behind their profile."
+    },
+    "text_settings_cover_image_description": {
+      "default": "Resetting restores the default artwork.",
+      "description": "Description of the settings row that lets users reset their custom cover image back to the default artwork."
+    },
+    "text_settings_no_cover_image_description": {
+      "default": "You have no cover image set.",
+      "description": "Description of the settings row that lets users reset their cover image, shown when the user has not set one and there is nothing to reset."
+    },
+    "button_text_reset_cover_image": {
+      "default": "Reset",
+      "description": "Text for the button that resets the user's custom cover image back to the default artwork. This must be as short as possible."
+    },
+    "button_label_reset_cover_image": {
+      "default": "Reset your cover image to the default artwork.",
+      "description": "Aria-label for the button that resets the user's custom cover image back to the default artwork."
+    },
+    "confirmation_title_reset_cover_image": {
+      "default": "Reset cover image?",
+      "description": "Title of the confirmation dialog shown before resetting the user's custom cover image."
+    },
+    "confirmation_message_reset_cover_image": {
+      "default": "This will remove your custom cover image and restore the default artwork. You can set a new one at any time.",
+      "description": "Body of the confirmation dialog shown before resetting the user's custom cover image."
+    },
     "switch_label_toggle_halloween_filters": {
       "default": "Toggle halloween filters",
       "description": "Aria-label for the switch that allows users to toggle halloween filters."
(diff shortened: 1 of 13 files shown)


seferturan (approved):
Applied some fixups:
- Removed some dead code/comments.
- Wrapped the reset in `try`/`finally`. A failed request used to leave the button disabled for good.
- The row no longer hides itself when there is nothing to reset. Button goes disabled and the description says so.
- Specs now cover enabled/disabled instead of the copy. Asserting on labels was too brittle.

PR #3134, opened 2026-08-14
feat(cover-image): preview the artwork before it becomes your cover

### LINKED ISSUE #2954: Show preview when setting cover image
Similar to android app

ElMagnea:
Closes #2954.
## Imagery
Screenshot 2026-08-14 at 18 06 50
Screenshot 2026-08-14 at 18 17 14

## What

Choosing **Set as cover image** on a movie, show or episode now opens a confirmation that shows the fanart it is about to use, so you decide against the picture instead of the title.

Before: the dropdown item fired the `PUT /users/set_cover` immediately, and the only way to see what you got was to go look at your profile.

## How

`ConfirmationDialog` learned one optional `previewUrl`, threaded through the confirmation layer (`mapToConfirmation` → `ConfirmationRequest` → provider → dialog) and rendered as a 16:9 plate between the message and the buttons, via `CrossOriginImage` so it fades in and keeps the environment-URI fallback. Any future confirmation that wants to show its subject can pass one; every existing dialog renders exactly as before.

`useCoverImage` now wraps the request in `useConfirm` — the same shape `useResetCoverImage` uses on the settings side.

The two call sites pass their own artwork: `media.cover.url.medium` for movies/shows, `episode.cover.url` for episodes. A media entry with no artwork still gets the confirmation, just text-only.

## Copy

`warning_prompt_set_cover_image` (\"Use \"{title}\" as your profile background image?\") already existed, was written for exactly this dialog, and was already translated into all 27 locales — it was simply never wired up. Only `confirmation_title_set_cover_image` is new, so this adds one string to translate.

## Verification

- \`deno fmt\` clean on touched paths, \`deno task check\` 0 errors / 0 warnings (7315 files), full \`vitest\` run 2804 passed.
- New \`SetCoverImageAction.spec.ts\`: opening the menu item shows the preview with the right \`src\` and fires no request; confirming sends \`{ cover_type: 'movie', cover_id: 1000 }\`.
- \`mapToConfirmation.spec.ts\`: affirmative operation, title interpolated, preview carried through, and the no-artwork case still confirms.
- Eyeballed in both themes.

## Notes

- **Stacked on \`feat/reset-cover-image\` (#3133)** — both PRs touch the same three confirmation files, so this sits on top rather than conflicting. Merge #3133 first and this retargets cleanly to \`main\`.
- Affirmative confirmations render their primary button in the same muted ghost treatment as the cancel button (pre-existing — \`RestoreShow\` and \`StartRewatching\` look the same). Left alone here since the button system is due its own pass, but it is the one thing I would change next.

### DIFF
--- a/.agents/rules/components.md
+++ b/.agents/rules/components.md
@@ -156,6 +156,25 @@ Use `{#snippet}` and `Snippet` type for flexible content injection.
 
 ---
 
+## Extending a Primitive: Slots, Not Domain Props
+
+`lib/components` primitives stay domain free when **modified**, not just when
+placed. If a feature needs one to render something new, add a `Snippet` slot and
+let the feature fill it, along with any styling specific to that content.
+
+```svelte
+<!-- Bad - the primitive now knows about one feature's data -->
+const { previewUrl }: { previewUrl?: HttpsUrl | Nil } = $props();
+
+<!-- Good - it renders whatever it is handed -->
+const { content }: { content?: Snippet } = $props();
+```
+
+Still props: generic options (`size`, `variant`, `aria-label`) and anything that
+gates behaviour rather than renders.
+
+---
+
 ## Provider Pattern
 
 Feature providers are thin shells calling a context factory from `_internal/`.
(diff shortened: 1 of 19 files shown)


seferturan (approved):
Applied some fixups/added some things:

- The confirmation dialog no longer knows what it is showing. It takes a `content` snippet and renders it, so `previewUrl` is gone from `Confirmation`, `ConfirmationRequest` and the dialog props. It is now more generic.
- `Confirmation` is now the single source for the fields that were declared three times over.
- `previewUrl` is typed `HttpsUrl`, both call sites already supply one.
- Reused the labels android already ships for this dialog instead of adding new keys.
- `docs`: updated the instructions to keep components more generic when adding new props.

PR #3135, opened 2026-08-14
fix(dialogs): give confirmation actions a visible edge

ElMagnea:
Stacked on #3134 (which is stacked on #3133). Review the last commit only — `19a2e1ce0`.

## The problem

Confirmation pop-ups ask you to commit to something, then render the two choices as near-invisible slabs. The recessed action is *darker* than the dialog it sits on — `shade-950` on `shade-930` in dark, `shade-60` on `shade-10` in light — so three to five percent of lightness is all that separates a button from the surface behind it.

Pulling on that thread turned up two more:

- **Affirmative dialogs render both actions identically.** `Set cover image?`, `Restore show?` and `Start rewatching?` hand both buttons the same `ghost` style, the same `custom` colour and the same variant. There was nothing distinguishing confirm from cancel beyond the words inside them.
- **Hover was broken in every confirmation dialog.** `--color-background-custom` sat on the shared action row, and `Button.svelte`'s ghost-hover rule (`[data-style=ghost]:hover:not(…):not(…)[data-variant=primary]`, specificity 0,6,0) outranks anything the dialog wrote at 0,5,0. Consequences: hovering **Cancel in a destructive dialog painted it the destructive red**, and in an affirmative dialog the token was undefined, so the rule resolved to an invalid colour and **the button vanished on hover**. `--color-background-button-outline` was invalid for the same reason, which quietly killed the focus ring.

## The fix

Everything lands in the shared `ConfirmationDialog`, so all ~35 `useConfirm` call sites inherit it.

- **Every action carries a rim.** Neutral (`--color-confirmation-button-border`) on the recessed ones; on filled ones it is mixed from their own label colour, so red and blue get an edge that belongs to them rather than a grey ring stapled on. Implemented as `border`, not an inset shadow, so `flat`'s hover glow survives untouched.
- **The affirmative confirm outranks its cancel** via the accent rim plus the primary label colour — no new fill colour introduced.
- **Colour tokens moved onto per-action slots** (`display: contents`), which is what fixes the red-cancel and vanishing-cancel bugs and restores the focus ring. The lesson generalises: don't fight `Button.svelte` on specificity from a consumer — hand it `--color-background-custom` / `--color-foreground-custom` and let its own rules paint.
- **Pointer feedback rides the rim**, not the fill, since a recessed surface has nowhere left to travel in either theme.

Two new token pairs in both `modes.scss` mixins; no component-level theme conditionals.

## Verification

`deno fmt` on the touched files, `svelte-check` clean (0 errors), full vitest green (2804 passed).

Eyeballed headlessly across the whole matrix — affirmative / destructive / preventative / challenge-disabled × light + dark × rest, hover and keyboard focus — through a throwaway design-system route that was removed before committing.

## Deliberately out of scope

- `ProfileImageCropDialog` and `ProfileImageViewDialog` still use bare header ghost buttons; same treatment would suit them.
- The affirmative confirm could go further and become a filled primary rather than a stroked one. That is a bigger swing than this fix, and the `/* FIXME: make this design leading for the new button styles */` in this file is still waiting on the wider button pass.

### DIFF
--- a/.agents/rules/components.md
+++ b/.agents/rules/components.md
@@ -175,6 +175,29 @@ gates behaviour rather than renders.
 
 ---
 
+## Restyling a Shared Component
+
+Do not restyle a shared component from its consumer. Use its props first
+(`style`, `variant`, `color`, `size`), then its documented CSS variable hooks.
+If the look you need does not exist, add it to the component so every caller
+gets it, instead of to one call site.
+
+```scss
+/* Bad - reimplements a stroke the button already ships as style="outline" */
+.my-row :global(.trakt-button) {
+  border: var(--border-thickness-xxs) solid var(--my-border);
+}
+```
+
+The tell is duplication: if the override reaches for the same tokens, or repeats
+the hover and disabled rules the component already has, the component already
+does it.
+
+One-off surfaces (landing, marketing pages) can override locally. Anything
+reused across the app should not.
+
+---
+
 ## Provider Pattern
 
 Feature providers are thin shells calling a context factory from `_internal/`.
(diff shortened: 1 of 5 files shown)


seferturan (approved):
Applied some fixups/fixes:

- The preview of the cover image now has a subtle shadow.
- The dialog actions use the button's own `style="outline"` instead of a hand rolled border. The stroke, hover tint, muted disabled rim and press feedback all already shipped with it. I.e. use the existing Button component as much as possible, without such a big local style override.
  - `outline`'s default stroke drops from 2px to 1px. Select triggers and most other borders already draw at 1px, so the button was the outlier. Checked the other 3 consumers, none of them override the stroke.
- `docs`: components.md now says shared components get styled through their props and variable hooks, not `:global()` overrides from the consumer.

PR #3147, opened 2026-08-17
Comment: watched percentage and count

### LINKED ISSUE #3137: Bring Back Watched Context for Reviews
### Code of Conduct

Often times people write reviews for shows after seeing just a couple of episodes. Then those reviews stay around regardless of how later episodes or seasons were received and even regardless of the reviewer kept watching.
Could be someone left a bad review after watching just a few episodes and didn't stick around for a potentially satisfying conclusion that would have tied things up nicely.
Could be someone left a good review after watching one or two seasons but didn't stick around afterwards when it all went downhill.
Similarly with movies, people leave comments/reviews having never seen them, even before they were released and nobody could see them.

v2 addressed this by adding a quick info how much of a show was watched (or for movies, if it was watched at all) to reviews. v3 doesn't currently have this.

I believe v2's info was based on how much has been watched to date even after the review was written, but probably more relevant would be how much had been watched when the review was written or last edited.

This info should be added to reviews to see how much of a basis there was for the review or how many episodes/seasons it even relates to for shows.

Ideally, there should also be a filter to filter by reviews where 100% of the current media has been watched when the review was written or last edited.

_No response_

### Additional context

_No response_

rectifyer:
# What

Add watched context on comments and reviews (#3137). Every comment card now shows a single stats pill in the top right, using the same pill format as the profile activity rows. Watched context reads first — an eye icon with the watched percentage for shows and seasons, or a play icon with the bare play count for movies and episodes — followed by a `·` separator and the commenter's star rating at pill scale. The pill smartly renders both segments, either one alone, or nothing at all when the commenter has no watch history and no rating. Hovering shows a tooltip with the exact context ("Watched 45 of 66 episodes" for shows/seasons, "2 plays" for movies/episodes), and the pill is hidden on your own comments.

# How

The data was already on every comment response (`user_stats.play_count` / `user_stats.completed_count` / `user_stats.rating`) and mapped into `MediaComment.user.stats`, it was just never rendered. The new `ReviewerStatsTag` component replaces the old standalone `CommenterRating` and renders everything through the existing `StemTag`, so the shell, type scale, and icon sizing match the activity pills for free. The percentage comes from `completed_count` against the aired episode count: `ShowEntry.episode.count` for shows, and the season's `episodes.count` for season comments, which is now plumbed through `SeasonCommentProps` (verified against the live API that `completed_count` is scoped to the commented item). The rating reuses `toUserRating` (5-star scale) with `StarIcon`, matching the activity pills.

Play counts display as a bare number; the wording lives in the tooltip and `aria-label` via `TagIntlProvider.toPlayCount`, which is now singular-aware via a new `tag_text_play` key, so `PlaysTag` elsewhere stops saying "1 plays" too. A new static `EyeIcon` joins the existing `PlayArrowIcon`; no new theme tokens were needed since the pill uses the standard stem tag colors.

# Screenshots

#### Comment

image

image

#### Movie

image

#### Show

image

#### Season

image

#### Episode

image

# Out of scope

The issue's stretch goals need API support and are not included: stats snapshotted at write/edit time (`user_stats` reflects current progress, same as v2), and a server-side "100% watched" comment filter.


seferturan (changes requested):
Applied some fixups:

- Rescoped from `comment` to `review`
- Folded `watchedEpisodes` into `episodeProgress`. Dropped the `isEpisodic` and `Number.isFinite` guards, both unreachable.
- Typed `media` as `MediaEntry | ShowEntry` so the `in` check narrows on its own. Drops the `as ShowEntry` cast.
- `SeasonsDrawerHost` was running `seasons.find()` twice. `currentSeasonData` already carries the id.
- `cursor: default` and `user-select: none` on the tag, it was showing a text cursor.

Did not approve yet, because it can get very cramped when all data is there 😅
Examples:
Image

Image

I'll also have a crack at it 👌

seferturan (approved):
Ended up feature flagging it for now: showing percentages here while we don't show it anywhere else seemed a bit strange. Maybe we show this info in the same way as the tags on the posters? 🤔

PR #3149, opened 2026-08-18
ci(share): prunes stale share images on a schedule

seferturan:
## 🎶 Notes 🎶

- Adds a daily job that prunes stale share images so cards get regenerated as their media changes
- Tiers mirror `AGGRESSIVE_RECHECK_SCHEDULE` in trakt-workers, so the two stay in step:
  - first 14 days since the anchor: recheck every 12h
  - 15-40 days: 3d
  - 41-120 days: 7d
  - after that: 30d
- Scoped to `images/share/{og,feed,story}/{movie,show}/`, derived from the same helper the endpoint builds paths with, so the prefixes cannot drift from where images actually land
- Reads `cachedAt` and `releasedAt` from object metadata. An image goes when its `cachedAt` age passes the recheck interval for its `releasedAt` age
- `releasedAt` is now the last aired episode for shows, the release date for movies
  - `first_aired` was wrong here. A long running show that premiered years ago sat in the slowest tier while it was actively airing and churning weekly
  - `updated_at` was the other candidate, but it re-stamps on every metadata refresh, and it is frozen into R2 at generate time, so it would pin anything the refresh pipeline touches to the 12h tier forever
  - needs `last_aired` on the show payload, added in https://github.com/trakt/trakt-workers/pull/1476 (merged and deployed)
- `last_aired` is not on `ShowResponse` in @trakt/api yet, so `mapToShowEntry` reads it through a module-private type extension for now. Drops out when the SDK catches up

## 🧪 Dry run 🧪

Ran it on CI against walter with the delete path stubbed out: https://github.com/trakt/trakt-web/actions/runs/32983215775

- All 6 prefixes resolved
- Would delete 389 of 5239 objects. 196 in the 12h tier, 88 in the 3d, 105 in the 7d
- Zero legacy entries, so nothing needed a one time reset
- Deleted exactly one candidate for real (`images/share/og/show/lanterns/image.png`, the oldest `cachedAt`) and HEADed it after to confirm it was gone, so the signed DELETE is proven too
- ⚠️ Those 389 were judged on the old `first_aired` anchors, since walter only gets `lastAired` based ones once this deploys. The temp commit that did all this has been reverted

## 🛠 Fixed along the way 🛠

- The egress allowlist would have failed on the first run. It only permitted `api.cloudflare.com`, but the script talks to `<account>.r2.cloudflarestorage.com` and `setup-deno` needs jsr and npm. Copied the proven set from `supply_chain.yml` and added the R2 host
- Kept the checkout before `setup-deno`. Dropped it at first thinking the composite action's own checkout covered it, but a local action cannot resolve without the repo on disk first, so CI died on `Can't find 'action.yml'` 😅
- Split `buildTargetPrefixes` and `shareImagePrefix` out of `buildImagePath` so each file keeps one export
- Spec fixtures were still on the pre move `images/og/...` prefixes
- Folded the run script onto `R2Client` in `.scripts/_internal/r2.ts` rather than keeping its own SigV4 signer, list and delete. Added `listPage`, `headMetadata` and a single key `delete` to the client, and put the existing `list()` generator on top of `listPage`. The run script drops from 266 lines to 68
  - kept a single key `DELETE` instead of reusing `deleteMany`, since the batch API answers 200 with per key errors in the body and would have inflated the deleted count and hidden failures from the exit code
  - `R2Client.list()` now builds its query sorted by name and %-encoded, which is what SigV4 specifies. Signing and the request use the same string either way, so signatures stay valid, but the immutable sync and prune scripts share that path 👀
- Runs as `deno task prune:share-images` from `projects/client`, matching how `ci_cd.yml` invokes the other two R2 scripts
- Spec durations go through `time.hours/days` instead of raw millisecond arithmetic

## 🤔 Worth knowing 🤔

- One HEAD request per object per run, since the S3 list API does not return custom metadata. Chunked 50 at a time
- Real media images in walter carry no `cachedAt`, so they are skipped even if a prefix were ever wrong. That is the safety net, not the plan
- The anchor migrates gradually. A stored `releasedAt` only becomes `lastAired` based when that image is regenerated, and this job is what triggers the regeneration, so the two phase in together
- The cron arms at merge, not at deploy. GitHub only runs `schedule` from the default branch, and the job talks to R2 directly rather than through the worker

### DIFF
--- a/.github/workflows/prune_share_images.yml
+++ b/.github/workflows/prune_share_images.yml
@@ -1,0 +1,49 @@
+name: The Purge of Stale Visions aka Prune Share Images
+
+on:
+  schedule:
+    - cron: '0 3 * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  prune:
+    name: Expunging the Stale Relics
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    permissions:
+      contents: read
+
+    steps:
+      - name: Sealing the Outbound Gates aka Harden Runner
+        uses: step-security/harden-runner@v2
+        with:
+          egress-policy: block
+          allowed-endpoints: >
+            api.github.com:443
+            codeload.github.com:443
+            deno.com:443
+            dl.deno.land:443
+            github.com:443
+            jsr.io:443
+            npm.jsr.io:443
+            registry.npmjs.org:443
+            release-assets.githubusercontent.com:443
+            *.r2.cloudflarestorage.com:443
+
+      - name: Secure the Evidence aka Checkout Code
+        uses: actions/checkout@v5
+
+      - name: Setup Deno Environment
+        uses: ./.github/actions/setup-deno
+
+      - name: Purge the Stale Visions aka Prune Share Images
+        working-directory: projects/client
+        run: 'deno task prune:share-images'
+        env:
+          CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+          R2_ACCESS_KEY_ID: ${{ secrets.R2_ACCESS_KEY_ID }}
+          R2_SECRET_ACCESS_KEY: ${{ secrets.R2_SECRET_ACCESS_KEY }}
+          R2_BUCKET_NAME: walter
--- a/projects/client/.scripts/pruneShareImages.run.ts
+++ b/projects/client/.scripts/pruneShareImages.run.ts
@@ -1,0 +1,67 @@
+import { buildTargetPrefixes } from '../src/routes/api/shareable-image/_internal/buildTargetPrefixes.ts';
+import { R2Client } from './_internal/r2.ts';
+import { pruneShareImages } from './pruneShareImages.ts';
+
+// A full 1000-key page of HEAD requests exhausts sockets and trips rate limits.
+const HEAD_CHUNK_SIZE = 50;
+
+function requireEnv(name: string): string {
+  const value = Deno.env.get(name);
+
+  if (!value) {
+    console.error(`Missing required env var: ${name}`);
+    Deno.exit(1);
+  }
+
+  return value;
+}
+
+const r2 = new R2Client({
+  accountId: requireEnv('CLOUDFLARE_ACCOUNT_ID'),
+  accessKey: requireEnv('R2_ACCESS_KEY_ID'),
+  secretKey: requireEnv('R2_SECRET_ACCESS_KEY'),
+  bucket: requireEnv('R2_BUCKET_NAME'),
+});
+
+// The S3 list API omits custom metadata, so every key needs its own HEAD.
+async function hydrateMetadata(keys: ReadonlyArray<string>) {
+  const objects = [];
+
+  for (let i = 0; i < keys.length; i += HEAD_CHUNK_SIZE) {
+    const hydrated = await Promise.all(
+      keys.slice(i, i + HEAD_CHUNK_SIZE).map(async (key) => ({
+        key,
+        customMetadata: await r2.headMetadata(key),
+      })),
+    );
+    objects.push(...hydrated);
+  }
+
+  return objects;
+}
+
+const bucket = {
+  list: async (opts?: { cursor?: string; prefix?: string }) => {
+    const page = await r2.listPage({
+      prefix: opts?.prefix,
+      cursor: opts?.cursor,
+    });
+
+    return {
+      objects: await hydrateMetadata(page.keys),
+      truncated: page.cursor !== undefined,
+      cursor: page.cursor,
+    };
+  },
+  delete: (key: string) => r2.delete(key),
+};
+
+const result = await pruneShareImages(bucket, buildTargetPrefixes());
+
+console.log(
+  `Prune complete. deleted: ${result.deleted}, skipped: ${result.skipped}, errors: ${result.errors}`,
+);
+
+if (result.errors > 0) {
+  Deno.exit(1);
+}
--- a/projects/client/.scripts/pruneShareImages.spec.ts
+++ b/projects/client/.scripts/pruneShareImages.spec.ts
@@ -1,0 +1,402 @@
+import { time } from '$lib/utils/timing/time.ts';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { pruneShareImages } from './pruneShareImages.ts';
+
+const NOW = new Date('2026-06-04T12:00:00Z').getTime();
+
+type BucketObject = {
+  key: string;
+  customMetadata?: Record<string, string>;
+};
+
+function makeBucket(objects: BucketObject[]) {
+  const deleted: string[] = [];
+
+  const bucket = {
+    list: vi.fn((opts?: { prefix?: string; cursor?: string }) => {
+      const prefix = opts?.prefix;
+      const filtered = prefix
+        ? objects.filter((o) => o.key.startsWith(prefix))
+        : objects;
+      return Promise.resolve({
+        objects: filtered,
+        truncated: false,
+        cursor: undefined,
+      });
+    }),
+    delete: vi.fn((key: string) => {
+      deleted.push(key);
+      return Promise.resolve();
+    }),
+    deleted,
+  };
+
+  return bucket;
+}
+
+function cachedAt(msBefore: number): string {
+  return new Date(NOW - msBefore).toISOString();
+}
+
+function releasedAt(daysAgo: number): string {
+  return new Date(NOW - daysAgo * time.days(1)).toISOString();
+}
+
+const prefixes = ['images/share/og/movie/', 'images/share/og/show/'];
+
+describe('pruneShareImages', () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+    vi.setSystemTime(NOW);
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+  describe('staleness tiers', () => {
+    it('prunes a fresh release cached over 12 hours ago (tier 1: ≤14 days)', async () => {
+      const bucket = makeBucket([
+        {
+          key: 'images/share/og/movie/the-dark-knight/image.png',
+          customMetadata: {
+            cachedAt: cachedAt(time.hours(13)),
+            releasedAt: releasedAt(7),
+          },
+        },
+      ]);
+
+      const result = await pruneShareImages(bucket, prefixes);
+
+      expect(result.deleted).toBe(1);
+      expect(result.skipped).toBe(0);
+    });
+
+    it('skips a fresh release cached under 12 hours ago (tier 1: ≤14 days)', async () => {
+      const bucket = makeBucket([
+        {
+          key: 'images/share/og/movie/the-dark-knight/image.png',
+          customMetadata: {
+            cachedAt: cachedAt(time.hours(11)),
+            releasedAt: releasedAt(7),
+          },
+        },
+      ]);
+
+      const result = await pruneShareImages(bucket, prefixes);
+
+      expect(result.deleted).toBe(0);
+      expect(result.skipped).toBe(1);
+    });
+
+    it('prunes a mid-age release cached over 3 days ago (tier 2: 15-40 days)', async () => {
+      const bucket = makeBucket([
+        {
+          key: 'images/share/og/movie/oppenheimer/image.png',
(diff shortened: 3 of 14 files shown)


projects/client/.scripts/pruneShareImages.run.ts
```diff
@@ -0,0 +1,272 @@
...
+}
+
+const accountId = requireEnv('CLOUDFLARE_ACCOUNT_ID');
+const accessKeyId = requireEnv('CLOUDFLARE_R2_ACCESS_KEY_ID');
+const secretAccessKey = requireEnv('CLOUDFLARE_R2_SECRET_ACCESS_KEY');
+const bucketName = requireEnv('R2_BUCKET_NAME');
+
+const endpoint = `https://${accountId}.r2.cloudflarestorage.com`;
+
+/**
```
rectifyer: **[blocking]** This duplicates `.scripts/_internal/r2.ts`, which already does SigV4, paginated list, and batched delete, and the copy drops behavior the proven client has:

- error messages omit the response body (`R2 list failed ...: 403` vs `R2Client`'s `LIST ... → 403 <body>`), which is what you need when debugging signature failures
- one DELETE per key instead of `deleteMany`'s multi-object batches (389 requests vs 1 on your dry-run numbers)
- no XML unescaping (see below)

I know the PR body defers the fold-in, but extending `R2Client` with a metadata-returning `head()` looks smaller than this parallel implementation. If it stays deferred, please at least port the unescaping now.

projects/client/.scripts/pruneShareImages.run.ts
```diff
@@ -0,0 +1,272 @@
...
+
+  if (!response.ok) {
+    throw new Error(
+      `R2 list failed for ${prefix ?? 'root'}: ${response.status}`,
+    );
+  }
+
+  const xml = await response.text();
+
+  const keys = [...xml.matchAll(/<Key>([^<]+)<\/Key>/g)]
```
rectifyer: **[blocking]** `<Key>` and `<NextContinuationToken>` values come back XML-escaped and are used raw here, while `R2Client.list` runs them through `xmlUnescape`. A key containing `&` would be parsed as `&amp;`, the DELETE would target the escaped key, S3 returns 204 for missing keys, and the run would report it "deleted" while the object survives. Slug validation (`/^[\w-]+$/`) makes this latent rather than live today, but it's a silent-wrong-answer class of bug and the fix already exists in `_internal/xml.ts`.

projects/client/.scripts/pruneShareImages.run.ts
```diff
@@ -0,0 +1,272 @@
...
+  if (!value) {
+    console.error(`Missing required env var: ${name}`);
+    Deno.exit(1);
+  }
+
+  return value;
+}
+
+const accountId = requireEnv('CLOUDFLARE_ACCOUNT_ID');
+const accessKeyId = requireEnv('CLOUDFLARE_R2_ACCESS_KEY_ID');
```
rectifyer: `r2FromEnv` in `_internal/r2.ts` reads `R2_ACCESS_KEY_ID` / `R2_SECRET_ACCESS_KEY`; this script introduces `CLOUDFLARE_R2_*` names for the same secrets. Two names for one credential across `.scripts` invites workflow copy-paste mistakes, and reusing the existing names would let this script use `r2FromEnv` after the planned fold-in.

projects/client/.scripts/pruneShareImages.run.ts
```diff
@@ -0,0 +1,272 @@
...
+  headers: Record<string, string>,
+  body: string,
+): Promise<string> {
+  const now = new Date();
+  const amzDate = `${
+    now.toISOString().replace(/[:-]|\.\d{3}/g, '').slice(0, 15)
+  }Z`;
+  const dateStamp = amzDate.slice(0, 8);
+
+  headers['x-amz-date'] = amzDate;
```
rectifyer: `signRequest` mutates its `headers` parameter to smuggle `x-amz-date` / `x-amz-content-sha256` back to the caller. `R2Client.sign` returns the headers instead; the mutation runs against our purity rule in code-principles.

projects/client/.scripts/pruneShareImages.run.ts
```diff
@@ -0,0 +1,272 @@
...
+    const raw = await r2List(opts?.cursor, opts?.prefix);
+
+    // Hydrate custom metadata for each object via HEAD in chunks
+    // to avoid socket exhaustion and rate limits (1000 per page is too bursty)
+    const chunkSize = 50;
+    const chunks = Array.from(
+      { length: Math.ceil(raw.objects.length / chunkSize) },
+      (_, i) => raw.objects.slice(i * chunkSize, (i + 1) * chunkSize),
+    );
+    const objects = await chunks.reduce(
```
rectifyer: The async `reduce` with spread accumulation is O(n²) copying and hard to read. A plain loop over chunks pushing into an array (or sequential `for await`) says the same thing more clearly; at ~5k objects either works, so this is about readability.

projects/client/.scripts/pruneShareImages.ts
```diff
@@ -0,0 +1,140 @@
...
+): Promise<PruneResult> {
+  const now = Date.now();
+  const result: PruneResult = { deleted: 0, skipped: 0, errors: 0 };
+
+  for (const prefix of prefixes) {
+    let cursor: string | undefined;
+
+    do {
+      const listed = await bucket.list({ cursor, limit: 1000, prefix });
+
```
rectifyer, line 77: `limit: 1000` is silently ignored: the `run.ts` adapter's `list` signature doesn't accept it. Harmless today since 1000 is the server default, but misleading. Either wire it through as `max-keys` or drop it from the call and the `R2Bucket` type.

projects/client/.scripts/pruneShareImages.spec.ts
```diff
@@ -0,0 +1,348 @@
...
+describe('pruneShareImages', () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+    vi.setSystemTime(NOW);
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+  describe('staleness tiers', () => {
```
rectifyer, line 56: Nice coverage of both sides of each threshold. Two boundary cases worth adding: `cachedAt` age exactly equal to the interval (deletes, since `isStale` uses `>=`) and `daysSinceRelease` exactly at a tier edge (14 vs just over 14).

projects/client/src/lib/requests/_internal/mapToShowEntry.ts
```diff
@@ -63,6 +67,7 @@ export function mapToShowEntry(
     airDate: effectiveReleaseDate,
     releaseDate: effectiveReleaseDate,
     effectiveReleaseDate,
+    ...(show.last_aired ? { lastAired: new Date(show.last_aired) } : {}),
```
rectifyer, line 70: The conditional spread diverges from this file's own idiom a few lines down (`updatedAt: show.updated_at ? new Date(show.updated_at) : undefined`). Since `lastAired` is `nullish()` in the schema, the plain ternary works and reads cleaner:

```suggestion
    lastAired: show.last_aired ? new Date(show.last_aired) : undefined,
```

.github/workflows/prune_share_images.yml
```diff
@@ -0,0 +1,47 @@
...
+    - cron: '0 3 * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  prune:
+    name: Expunging the Stale Relics
+    runs-on: ubuntu-latest
```
rectifyer, line 14: Two one-liners worth adding: `timeout-minutes` (this job makes ~5k network calls, a hang burns a 6h runner) and a `concurrency` group so a `workflow_dispatch` can't overlap the cron. Deletes are idempotent so the risk is small, but both are cheap.

rectifyer (changes requested):
Well-scoped and carefully validated, the tier logic and its tests are solid, and the dry run write-up is exemplary. One blocker before merge: the reimplemented S3 client in `pruneShareImages.run.ts` is weaker than the `R2Client` we already have in `.scripts/_internal/r2.ts` (missing XML unescaping, no response body in errors, unbatched deletes) while also being the only untested part of the PR. Either fold onto `R2Client` now or at least port the `xmlUnescape` calls, details inline.

One nit with no good inline anchor: every sibling in `.scripts/` is kebab-case (`prune-immutable-r2.ts`, `check-i18n-placeholders.ts`) while the new trio is camelCase (`pruneShareImages.ts`). Within this directory the new names are the odd ones out, consider renaming to `prune-share-images*.ts`.

projects/client/.scripts/pruneShareImages.run.ts
seferturan: Done 👌 Both `<Key>` and `<NextContinuationToken>` go through `xmlUnescape` now.

projects/client/.scripts/pruneShareImages.run.ts
seferturan: Done 👌 Uses `R2_ACCESS_KEY_ID` / `R2_SECRET_ACCESS_KEY` now. The workflow was already pointing at those secrets, so it was purely a rename in the env block.

projects/client/.scripts/pruneShareImages.run.ts
seferturan: Done 👌 It returns the signed headers now, and derives `host` itself so callers stop passing an object in at all.

projects/client/.scripts/pruneShareImages.run.ts
seferturan: Done 👌 Plain chunk loop, same shape as `deleteMany` in `_internal/r2.ts`.

projects/client/.scripts/pruneShareImages.ts
seferturan: Done 👌 Dropped it from the call and from the `R2Bucket` type.

projects/client/.scripts/pruneShareImages.spec.ts
seferturan: Done 👌 Added three: cached age exactly at the interval, release aged exactly 14d (stays tier 1), and 14d + 1h (drops to tier 2).

projects/client/.scripts/pruneShareImages.run.ts
seferturan: Ported the unescaping and the response bodies in the error messages now 👌

Full fold-in stays deferred: `R2Client.head()` returns a boolean and `list()` is an `AsyncGenerator<string>` with no cursor or metadata surface, so pruning needs it reshaped rather than swapped in. Will tackle that separately.

.github/workflows/prune_share_images.yml
seferturan: Added `timeout-minutes: 30` 👌

Left the concurrency group out for now. Deletes are idempotent, so the worst case on an overlap is an inflated count in the log line, and nothing else in `.github/workflows` uses one. Happy to add it if you feel strongly.

projects/client/src/lib/requests/_internal/mapToShowEntry.ts
seferturan: No, the conditional spread is deliberate here. The mapped mocks are compared with `to.deep.equal`, which distinguishes an absent key from an explicit `undefined`, so the ternary turns 24 tests across 21 files red:

```
- Expected
+ Received
+       "lastAired": undefined,
```

`updatedAt` only gets away with it because every show mock has `updated_at` set.

Also `last_aired` is not `nullish()` in the API schema, it is the local `ShowResponseWithLastAired` augmentation this PR adds. Leaving as is 😅

projects/client/.scripts/pruneShareImages.run.ts
vladjerca: Folded it in now, so the deferral note above is stale. `R2Client` picked up `listPage`, `headMetadata` and a single key `delete`, and `run.ts` is a thin adapter over it. 266 lines down to 68.

Kept one DELETE per key rather than `deleteMany` though: the batch API answers 200 with per key errors in the XML body, so a failed object would count as deleted and the job would still exit 0. Happy to batch it if you want the request count down, but it needs the response parsed first.

.github/workflows/prune_share_images.yml
vladjerca: `timeout-minutes: 30` is in. Agreed on leaving the concurrency group out, deletes are idempotent and no other workflow here uses one. Shout if you want it anyway 👌🏼

PR #3154, opened 2026-08-19
fix: avoid onload handler crash during svelte:head hydration

Marius-TV:
The compiled onload listener on the async Plyr CSS link caused Svelte's head hydration replay to call removeAttribute on a mismatched node, crashing with "e.removeAttribute is not a function". Using a plain string onload attribute avoids the compiled event/hydration-replay path entirely.

## Summary
- Hydration was crashing with `TypeError: e.removeAttribute is not a function` because Svelte's compiled `onload` handler on the async Plyr CSS `<link>` inside `<svelte:head>` triggers a hydration-replay path (`replay_events`) that calls `removeAttribute` on the matched DOM node — a preceding author comment in the head throws off Svelte's node-matching walk, so it resolves to the wrong node.
- Replaced the function-based `onload={...}` handler with the classic plain-string `onload="this.media='all'"` attribute, which Svelte emits as static HTML rather than a compiled JS listener, sidestepping the hydration-replay path entirely.

## Test plan
- [ ] Load the app in the browser and confirm no hydration errors in the console
- [ ] Verify Plyr CSS still swaps `media` from `print` to `all` once loaded (video player styling intact)


seferturan (approved):
Issue seems to have been caused by a combination of how Svelte works and what Cloudflare does. Svelte finds `<svelte:head>` elements by counting positions, not by name. Ours compiled to `sibling(first_child(fragment), 12)`. Cloudflare's Google Fonts optimisation rewrites our head at the edge: drops the two font preconnects, inlines the stylesheet. Everything after shifts by 3. So Svelte counts to 12, grabs a whitespace text node, and calls `removeAttribute` on it. Svelte caught the error and remounted the whole app client side, so every cold load was paying a full re-render.

I moved all the static head assets to `app.html`. `<svelte:head>` is just `<title>` now, so no node references and nothing to desync. Updated the instructions accordingly.

Reproduced locally with the rewrite simulated, will verify prod once it's deployed.

PR #3159, opened 2026-08-19
Person: fix nav links on person media pages

### LINKED ISSUE #3128: Using bottom menu bar when viewing a list of actor's movies leads to 404
### Code of Conduct

https://app.trakt.tv/people/meryl-streep/movies?movies=acting&mode=media

It seems that trying to tap any one of the buttons in the menu bar when viewing a list of movies leads to 404. Tapping on one of the movies instead works fine and I can tap on any buttons in the menu bar then.

### Additional information

https://github.com/user-attachments/assets/7d71be59-f815-4fe6-9423-c2f989cd3ee5

rectifyer:
# Summary

Fixes #3128. On a person credits page like `/people/meryl-streep/movies?movies=acting&mode=media`, tapping any navbar link (home, discover, lists, search) rendered the destination page as a 404. Tapping a movie first, then using the navbar, worked fine.

# Root cause

The 404 was never a routing failure, it was the global error page being triggered by a stray API call:

1. The person credits pages hand a `headerActions` snippet to `NavbarStateSetter`, which pushes it into a global store. `CreditsPositionSelector` is therefore rendered inside the persistent navbar, not the page.
2. When navigating away, the old page is torn down but the navbar keeps rendering that snippet for a beat. The selector's `slug` prop (closed over from the dead page's `params.slug`) flips to `undefined`, and `fromRune` pushes that into `useCreditsList`, which built a new query and fetched `GET /people//movies` (no slug), which the API answers with 404. It fired twice: once with the stale filter set, once after the navbar filter state reset.
3. `ErrorProvider` treats any API 404 as a page-level 404. Since the response lands after the navigation completes, it paints the freshly loaded destination page as "404: Nothingness."

This also explains the reporter's workaround: tapping a movie first replaces the navbar header state before the selector can misfire with a dead slug.

# Fix

Add the documented `enabled` guard (requests.md, "Conditional Queries") to both credits queries so a blank slug never reaches the network:

- `personMovieCreditsQuery`
- `personShowCreditsQuery`

This covers every consumer at once: `CreditsPaginatedList`, `CreditsList`, the navbar-hosted `CreditsPositionSelector`, and `useHistoryCreditsList`. The shows page (`/people/[slug]/shows`) had the identical exposure and is covered by the same guard.

# Verification

Reproduced with a headless browser against the dev server: before the fix, clicking home fired two `GET /api/trakt/people/movies` requests (404) roughly 400ms after the click and the landing/home page rendered as a 404. After the fix, no empty-slug requests fire and the destination page renders normally.

# Follow-up candidates (out of scope)

- `ErrorProvider` retroactively 404s the next page: any query that 404s after navigation completes (retries can land up to ~7s later) paints whatever page is now active as a 404. Scoping `NotFoundError` to the navigation it belongs to would kill this bug class.
- Navbar `headerActions` snippets outlive their page, so any snippet reading route params can hit the same window.


seferturan (approved):
I applied a fixup:

Disabling the query for a slug param that should always be defined at that level was hiding the root cause.

The position selector lives in the navbar, not the page, so it outlived the page and kept re-reading `page.params` after you navigated away, firing a credits request with the wrong (or empty) slug. Moved the query up into the page and left the dropdown dumb, so the subscription dies with the page and nothing can fire a bogus request 👌

PR #3160, opened 2026-08-19
Episode: make long episode titles readable

### LINKED ISSUE #3140: No Way to View Long Episode Titles
### Code of Conduct

v3 Desktop, e.g.: https://app.trakt.tv/shows/clerks?season=1&view=episode&episode=1

A long enough title is cut off in the episode section of the show page without a hover title, same as in the season drawer episode tab.
That's not good, but it would be less bad if the full title was at least visible in the episode details. Unfortunately it's not:

Image

There should be hover titles for cut off titles in the episode section and the season drawer cards, and the title should just be shown in full in the episode details.

### Additional information

_No response_

rectifyer:
# Problem

Long episode titles were cut off with no way to read them in full: the episode drawer header truncated to a single line with no hover title, and the episode cards on the show page and in the season drawer did the same. Closes #3140.

# Changes

**Episode drawer header**: the episode title now wraps up to 3 lines before truncating, and when it still overflows, hovering (or tapping on touch) shows the full title in a tooltip. The tooltip only activates when the text is actually clamped, measured via the existing `lineClamp` action. The close button stays pinned to the top right when the title wraps, and the header now respects the spoiler setting: the title portion gets the standard spoiler blur and the tooltip is disabled while hidden, with the `S1 • E1` label staying readable.

**Episode cards (show page, season drawer, episode drawer season tab)**: titles clamp as before (1 line, or 2 on desktop summary cards) but now show a truncation tooltip with the spoiler-safe title when cut off. `word-break: break-all` on single-line clamps keeps the mid-word ellipsis look of the old `text-overflow: ellipsis`, while multi-line titles keep word-boundary wrapping. No tooltip is shown while the spoiler hides the title.

# Screenshots

#### Episode Drawer

image

#### Episode Card with tooltip

image

# Implementation notes

The shared `Drawer` primitive is untouched: the episode drawer passes a `metaInfo` snippet through the existing `string | Snippet` prop. Tooltip enablement follows the `NavGroup` precedent (`disabled` bound to a `lineClamp` clamp signal). Clamped blocks are keyed on the rendered title so measurements re-run when the text changes, since `lineClamp` only re-measures on resize. Also removed a stray `//` comment that sat inside the `<Drawer` tag and was parsed as bogus attributes.


seferturan (approved):
Applied some fixups:
- Episode titles now stay on a single line and truncate with `text-overflow: ellipsis`, with a tooltip carrying the full title. Applies to `EpisodeCard`, `MediaSummaryCard` and the episode drawer header.
    - Single line in the drawer on purpose, so the header doesn't jump height when stepping between episodes.
- Tooltip only shows when the title actually got cut off. New `trackTextOverflow` action measures on mount and on `pointerenter`, no `-webkit-line-clamp` and no `{#key}` blocks needed.

And added a fix for a pre-existing issue:
- Blur wasn't reactive. Svelte binds an action's function once at mount, so the new observable from `$derived` was never picked up and the blur stayed stuck on whichever episode was opened first. The action now takes the observable as its argument and resubscribes on `update`.
  - `spoilMeAnyway` takes the variant for the same reason, which drops the `NOOP_FN` action swap here and in `CommentBody`.

Issue #2055, opened 2026-04-08
Importer updates


seferturan:
The first version of the importer will be feature flagged. To make it more usable and unfaggable:

- Add guidelines/descriptions.
- Add feedback when a file can't be parsed.
- More verbose result summary

PR #3163, opened 2026-08-20
fix(import): align json and csv import with our guidelines and export

seferturan:
## 🎶 Notes 🎶

Collection of fixes so the JSON and CSV importers support what our own guidelines document, and can ingest our own export.

**Guidelines**

- Flat `*_id` entries dropped `type`. A documented `"type": "show"` went out as a movie lookup against a TMDB show id, so no match, or a match on an unrelated movie.
- `type: "season"` was documented but never implemented. No branch in `toType`, no seasons bucket anywhere.
- Only movies had `tmdb` in their id priority, so a show or episode carrying just a `tmdb_id` resolved to nothing and was dropped.
- Episodes never resolved by `imdb`. They were pushed into the `shows` bucket under an episode imdb id, which no show carries, so they vanished or marked a whole show watched. Also fixes IMDb `tvEpisode` rows.
- Ratings and watchlist payloads had no `episodes` bucket, so episode ratings and watchlisted episodes were dropped.
- One entry can carry a watch, a rating and a watchlist date at once, like our own example does. `inferAction` only ever returned one of them. CSV already fanned these out, JSON now matches.
- String ids (`"tmdb_id": "67324"`) are coerced at the parse boundary.
  - Contract hygiene rather than a fix. Rails casts them and the bulk resolver keys by `to_s`, so they did resolve.

**Export**

- The zip reader looked for `watched/history*`, `ratings/ratings*` and `lists/watchlist.json`. Exports name their entries `watched-history-1.json`, `ratings-movies-1.json` and `lists-watchlist.json`, so nothing matched and a raw export imported zero items.
  - Separators are folded together rather than swapped, so the v2 layout keeps working for anyone holding an older export.
- Season ratings nest their ids under `season.ids`, sibling to a `show`. Those were falling through to the show's ids.

**Verified**

- Three real exports through the parser and payload builders: 22643, 44568 and 26445 items, all landing in a payload, none dropped.
- A reporter's 9408 row CSV, same. Before, only the 881 movies resolved.
- TV Time and Letterboxd unchanged. Diffed payloads against the base branch engine on real fixtures. IMDb does change, and both changes are the episode fix above.

**Follow ups**

- ⚠️ Seasons cannot resolve by imdb, there is no imdb branch for them server side. Guidelines still promise trakt/imdb/tmdb for every type, so that line needs a caveat.
- `@trakt/api` episode ids omit `tmdb` and `imdb`, and season `tmdb` is typed as a string. All wrong against `find_by_multi`, the payload builders cast around it for now.
- Silently dropped items are still invisible in the UI, left the fixme for #2055.

### DIFF
--- a/projects/client/src/lib/sections/settings/import/engine/buildHistoryPayload.ts
+++ b/projects/client/src/lib/sections/settings/import/engine/buildHistoryPayload.ts
@@ -2,55 +2,50 @@ import type { HistoryAddRequest } from '@trakt/api';
 import {
   DEFAULT_EPISODE_MATCH_MODE,
   type EpisodeMatchMode,
+  type ImportType,
   type UniversalImportItem,
 } from '../ImportTypes.ts';
 import {
-  EPISODE_IDS,
+  type IdPriority,
   MOVIE_IDS,
   pickIds,
+  type ResolvedIds,
   SEASON_IDS,
   SHOW_IDS,
+  toEpisodeIdPriority,
 } from './pickIds.ts';
 
-type HistoryMovie = NonNullable<HistoryAddRequest['movies']>[number];
-type HistoryShow = NonNullable<HistoryAddRequest['shows']>[number];
-type HistorySeason = NonNullable<HistoryAddRequest['seasons']>[number];
-type HistoryEpisode = NonNullable<HistoryAddRequest['episodes']>[number];
+type HistoryEntry =
+  | { ids: ResolvedIds; watched_at?: string }
+  | { title: string; year: number; watched_at?: string };
 
-// Movies never fall back to {title, year}: server-side text matching
-// is too fuzzy and mismatches pollute history. Unresolved movies are
-// dropped instead (resolveMovieIds runs before this).
-function toHistoryMovie(
+function toHistoryEntry(
   { ids, watched_at }: UniversalImportItem,
-): HistoryMovie | null {
-  const resolvedIds = pickIds(ids, MOVIE_IDS);
-  if (resolvedIds) return { ids: resolvedIds as never, watched_at };
+  priority: IdPriority,
+): HistoryEntry | null {
+  const resolvedIds = pickIds(ids, priority);
+  if (resolvedIds) return { ids: resolvedIds, watched_at };
   return null;
 }
 
-function toHistoryShow(
-  { ids, title, year, watched_at }: UniversalImportItem,
-): HistoryShow | null {
-  const resolvedIds = pickIds(ids, SHOW_IDS);
-  if (resolvedIds) return { ids: resolvedIds as never, watched_at };
+// Only shows fall back to {title, year}: server-side text matching is too fuzzy
+// for movies and mismatches pollute history, so unresolved movies are dropped
+// instead (resolveMovieIds runs before this).
+function toHistoryShow(item: UniversalImportItem): HistoryEntry | null {
+  const entry = toHistoryEntry(item, SHOW_IDS);
+  if (entry) return entry;
+
+  const { title, year, watched_at } = item;
   if (title && year) return { title, year, watched_at };
   return null;
 }
 
-function toHistorySeason(
-  { ids, watched_at }: UniversalImportItem,
-): HistorySeason | null {
-  const resolvedIds = pickIds(ids, SEASON_IDS);
-  if (resolvedIds) return { ids: resolvedIds as never, watched_at };
-  return null;
-}
-
-// Prefer the episode's own id (tvdb/trakt) over positional resolution: the
-// export's episode id is the exact identity of what was watched and survives
-// season/episode renumbering divergence between TVDB and Trakt. Positional
-// (show id + season/number) is the fallback for episodes carrying no own id.
+// Prefer the episode's own id over positional resolution: the export's episode
+// id is the exact identity of what was watched and survives season/episode
+// renumbering divergence between TVDB and Trakt. Positional (show id +
+// season/number) is the fallback for episodes carrying no own id.
 function hasEpisodeId(item: UniversalImportItem): boolean {
-  return pickIds(item.ids, EPISODE_IDS) != null;
+  return pickIds(item.ids, toEpisodeIdPriority(item)) != null;
 }
 
 function isPositional(
@@ -63,6 +58,10 @@ function isPositional(
 type PositionalEpisode = { number: number; watched_at?: string };
 type ShowIds = { tvdb?: number; imdb?: string };
 type ShowGroup = { ids: ShowIds; seasons: Map<number, PositionalEpisode[]> };
+type PositionalShow = {
+  ids: ShowIds;
+  seasons: Array<{ number: number; episodes: PositionalEpisode[] }>;
+};
 
 function toShowIds(item: UniversalImportItem): ShowIds {
   return {
@@ -80,7 +79,7 @@ function toPositionalShows(
   items: ReadonlyArray<
     UniversalImportItem & { season: number; episode: number }
   >,
-): HistoryShow[] {
+): PositionalShow[] {
   const byShow = items.reduce((shows, item) => {
     const key = item.showTvdb != null
       ? `tvdb:${item.showTvdb}`
@@ -101,11 +100,11 @@ function toPositionalShows(
   return [...byShow.values()].map(({ ids, seasons }) => ({
     ids,
     seasons: [...seasons].map(([number, episodes]) => ({ number, episodes })),
-  })) as unknown as HistoryShow[];
+  }));
 }
 
 export function buildHistoryPayload(
-  items: UniversalImportItem[],
+  items: ReadonlyArray<UniversalImportItem>,
   episodeMatch: EpisodeMatchMode = DEFAULT_EPISODE_MATCH_MODE,
 ): HistoryAddRequest {
   const episodeItems = items.filter((item) => item.type === 'episode');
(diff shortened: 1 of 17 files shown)


rudf0rd (commented):
looked through the json side of this.  a handful of questions inline, mostly around cases where an item gets counted as imported but never actually lands.  might be missing context on some of these so take them as questions.

projects/client/src/lib/sections/settings/import/parsers/TraktJsonParser.ts
```diff
@@ -127,79 +139,99 @@ function isMultiIdFlatEntry(entry: TraktJsonEntry): boolean {
...
+  return filename.replaceAll('/', '-');
+}
+
 function isRelevantJsonFile(filename: string): boolean {
-  return filename.startsWith('watched/history') ||
-    filename === 'lists/watchlist.json' ||
-    filename.startsWith('ratings/ratings');
+  const path = toComparablePath(filename);
+  return path.startsWith('watched-history') ||
+    path === 'lists-watchlist.json' ||
```
rudf0rd: history and ratings use startsWith but watchlist is an exact match.  the export splits at 250 per page, so anyone with 251+ on their watchlist gets `lists-watchlist-1.json` / `-2.json` and no plain `lists-watchlist.json` at all.  looks like those would get filtered out here before we ever parse them.  should this be `startsWith('lists-watchlist')` too?  same in inferActionFromPath below.

projects/client/src/lib/sections/settings/import/parsers/TraktJsonParser.ts
```diff
@@ -127,79 +139,99 @@ function isMultiIdFlatEntry(entry: TraktJsonEntry): boolean {
...
+    }];
+  }
 
-  const media = type === 'episode' ? entry.show : (entry.movie ?? entry.show);
-  const episodeData = type === 'episode' ? entry.episode : undefined;
+  const items: UniversalImportItem[] = [
+    ...watchedAt
+      ? [{ ...base, action: 'history' as const, watched_at: watchedAt }]
+      : [],
+    ...entry.rating != null
```
rudf0rd: `!= null` lets `rating: 0` through, and clampRating downstream bumps 0 to 1.  a third party dump that writes `rating: 0` as "unrated" on every watched row would end up rating the whole history 1/10.  csv already guards NaN, is there a reason not to do `rating > 0` here?  or drop out of range instead of clamping?

projects/client/src/lib/sections/settings/import/engine/pickIds.ts
```diff
@@ -3,13 +3,16 @@ import type { ImportIds } from '../ImportTypes.ts';
 export type IdPriority = ReadonlyArray<keyof ImportIds>;
 
 export const MOVIE_IDS: IdPriority = ['imdb', 'tmdb', 'trakt'];
-export const SHOW_IDS: IdPriority = ['imdb', 'tvdb', 'trakt'];
-export const EPISODE_IDS: IdPriority = ['tvdb', 'trakt'];
+export const SHOW_IDS: IdPriority = ['imdb', 'tvdb', 'tmdb', 'trakt'];
+export const SEASON_IDS: IdPriority = ['tvdb', 'tmdb', 'trakt'];
+export const EPISODE_IDS: IdPriority = ['tvdb', 'tmdb', 'trakt', 'imdb'];
```
rudf0rd: two things i want to make sure i'm reading right:

1. tmdb is ahead of trakt now.  an export episode always has `ids.trakt`, so if tvdb is null we'd send the tmdb id and rely on the mapping instead of the exact id we already have.  was that intentional?  feels like trakt should win whenever it's present.
2. `episodeIdsRequestSchema` in @trakt/api is trakt|tvdb only.  the `as HistoryAddRequest` cast in the builders means tsc won't complain about tmdb/imdb here.  does the server actually resolve episodes by those?

projects/client/src/lib/sections/settings/import/engine/buildHistoryPayload.ts
```diff
@@ -107,30 +121,24 @@ export function buildHistoryPayload(
...
-    ...leftoverEpisodes.flatMap(({ ids, watched_at }) =>
-      ids.imdb ? [{ ids: { imdb: ids.imdb } as never, watched_at }] : []
+  return {
+    movies: collect('movie', (item) => toHistoryEntry(item, MOVIE_IDS)),
+    shows: [
+      ...collect('show', toHistoryShow),
+      ...toPositionalShows(positionalEpisodes),
+    ],
+    seasons: collect('season', (item) => toHistoryEntry(item, SEASON_IDS)),
+    episodes: idEpisodes.flatMap((item) =>
```
rudf0rd, line 140: with EPISODE_IDS widened to tmdb/imdb, i think csv and liberator episode rows change behavior.  those carry the *show's* id in `ids` and never set showTvdb/showImdb, so isPositional is false and they now go out as `episodes:[{ids:{tmdb:<show id>}}]`.  main dropped them.  e.g. `tmdb_id=1396,type=episode,season=1,episode=1` would send tmdb 1396 (breaking bad the show) as an episode id.  am i reading that right?  if so maybe keep EPISODE_IDS to trakt|tvdb, or gate the wider set on the id actually being an episode id.

projects/client/src/lib/sections/settings/import/engine/buildHistoryPayload.ts
```diff
@@ -107,30 +121,24 @@ export function buildHistoryPayload(
...
-    ...toPositionalShows(positionalEpisodes),
-    ...leftoverEpisodes.flatMap(({ ids, watched_at }) =>
-      ids.imdb ? [{ ids: { imdb: ids.imdb } as never, watched_at }] : []
+  return {
+    movies: collect('movie', (item) => toHistoryEntry(item, MOVIE_IDS)),
+    shows: [
+      ...collect('show', toHistoryShow),
+      ...toPositionalShows(positionalEpisodes),
+    ],
+    seasons: collect('season', (item) => toHistoryEntry(item, SEASON_IDS)),
```
rudf0rd, line 139: SEASON_IDS has no imdb, but isValidItem accepts an imdb-only season (guide still says imdb works for every type).  looks like that item passes review, ends up in an empty `seasons: []`, still gets counted by processChunks, and never shows in unresolved since that filter is movie-only.  so the user sees 1 success and nothing imported.  worth either rejecting imdb-only seasons up front or widening the unresolved filter?  not sure which you'd prefer.

projects/client/src/lib/sections/settings/import/parsers/TraktJsonParser.ts
seferturan: Good catch, and this one's bad. `startsWith('lists-watchlist')` in both spots now.

Checked it against a real export: 5 pages, 1215 items, every one of them filtered out before parsing. Total went 44568 -> 45783 after the fix.

I had that 0 in front of me while verifying and read it as "this export has no watchlist" 😅

Done 👌

projects/client/src/lib/sections/settings/import/parsers/TraktJsonParser.ts
seferturan: Fair. Dropping anything below 1 now instead of clamping up, still clamping above 10.

Worth flagging there was an explicit spec for the old behaviour (`should clamp ratings below 1 to 1`, with 0 as the example), so it was a deliberate call at some point. Rewrote it. If we'd rather read 0 as "lowest" instead of "unrated" I'll put it back.

Small correction on the csv side: it guards NaN but not 0, so `"0"` took the same path there too.

projects/client/src/lib/sections/settings/import/engine/pickIds.ts
seferturan: 1. Not intentional, that's a slip on my side. I put tmdb ahead of trakt, and the commit message even claims the opposite ("inserted ahead of trakt only, so anything that already resolved keeps resolving to the same id"). Reordered so trakt wins whenever it's there, same for shows and seasons.

2. Yes, it does. `Episode.find_by_multi` resolves trakt -> tmdb -> tvdb -> imdb and `BULK_SYNC_ID_COLUMNS` matches it. `episodeIdsRequestSchema` is just wrong, it omits both. Season `tmdb` is typed as a string there too. Noted as a follow up to fix upstream.

Done 👌

projects/client/src/lib/sections/settings/import/engine/buildHistoryPayload.ts
seferturan: You're reading it right. Went with your second option.

Couldn't do the first one: `TvTimeNativeParser` sets `ids: { tvdb: episode_id }` (the episode's own id) alongside season/episode and no show ids, so gating on "has positional numbers" would have broken tv time.

So `toEpisodeIdPriority` keeps trakt|tvdb when season+episode are present and uses the wide set otherwise. Your tmdb 1396 case resolves to nothing now rather than a wrong episode, and the reporter's csv is unaffected since it has no season/episode columns at all.

Doesn't actually fix liberator though, worth being clear about that. Its episode rows carry the show's tvdb in `ids` and main already sent those as episode ids. Proper fix is those parsers setting showTvdb/showImdb so positional resolution kicks in, which needs a showTmdb field too. Separate PR.

projects/client/src/lib/sections/settings/import/engine/buildHistoryPayload.ts
seferturan: Went with rejecting up front. `isValidItem` checks the ids against the type's own priority now, so an imdb-only season never makes it past parsing or into the count.

Surfacing dropped items properly in the UI is the bigger one, left the fixme for #2055.

Done 👌

rudf0rd (approved):

PR #3164, opened 2026-08-20
List: link to lists from the manage lists drawer

### LINKED ISSUE #3105: Link to lists in "Manage Lists" drawer
### Code of Conduct

Sometimes you want to immediately navigate to a list after assigning a movie to it from the "Manage List" drawer, e.g. to adjust the ranking or look up others in that list. Or you want to have a look into some list candidates in a new tab right before assigning the movie.
What you don't want to do is go the long way around, close the drawer, go to your list view, then the extra click to get to your personal list, then find it and click it. Yet this is was v3 currently requires you do.
v2 had links to all lists directly in its list management popup so you could easily and immediately reach all lists from there.

v3 should add links to the lists in the "Manage Lists" drawer as well.
Right now the entire item is the button to change the assignment. I could see it being a split button instead where there is a small portion of the button with an icon on the left - or the usual ">" caret on the right - as a link with the rest of the button retaining its current functionality.

_No response_

### Additional context

_No response_

rectifyer:
# Summary

Adds direct navigation to lists from the **Manage lists** drawer (closes #3105). Each row is now a split control: the main area still toggles membership, and a new divided caret segment on the trailing edge links to the list page, with a "View list" tooltip on hover. The watchlist row gets the same treatment, linking to `/users/me/watchlist`.

As a follow-up sprinkle, adding or removing an item now flashes the row background using the same effect as list reordering: purple when added, muted red when removed.

# Changes

- `DropdownItem` gains an `action` snippet rendered as a divided trailing segment whose pointer/focus events never reach the row, with its own hover, active, and focus-ring styling, plus a `flash` variant (`data-flash`) that overlays the purple/red background animation.
- `ListDropdownItem` builds the list URL from `ownerSlug`/`slug` with a numeric-id fallback, and the drawer's rows read canonical slugs overlaid by `useAllPersonalLists` from the personal + collaboration list summary queries (the minimal v3 payload carries no slugs).
- `WatchlistButton` forwards `action` and `flash` to its dropdown-item rendering.
- `ListsDrawer` flashes rows off the listed-ids diff (survives the re-sort after toggling) and the watchlist row off the update lifecycle; the row layout switches from `row-reverse` to bookmark-leading with the caret trailing.
- The reorder flash is now shared: `background-flash` keyframes + `--animation-duration-background-flash` token in `style/animations`, a `useBackgroundFlash` rune hook, and `BACKGROUND_FLASH_DURATION`; `ReorderDrawer` consumes all three instead of its local copies.
- New i18n keys: `link_label_view_list`, `tooltip_view_list`, `link_label_view_watchlist`, `tooltip_view_watchlist`.

# Screenshots

#### View list icon + tooltip
image

#### Highlight after adding to a list
Xnapper-2026-08-20-13 20 29

#### What it looked like before
before

# Notes

- Reduced motion is respected for free: the flash duration token derives from `--transition-increment`, which collapses to `0ms` under `prefers-reduced-motion`.
- Cancelling the remove confirmation never flashes, since the listed-ids never change.

### DIFF
--- a/projects/client/i18n/meta/en.json
+++ b/projects/client/i18n/meta/en.json
@@ -2444,6 +2444,27 @@
         }
       }
     },
+    "link_label_view_list": {
+      "default": "View list {name}",
+      "description": "Aria-label for the link that opens a list from the manage lists drawer.",
+      "variables": {
+        "name": {
+          "type": "string"
+        }
+      }
+    },
+    "tooltip_view_list": {
+      "default": "View list",
+      "description": "Tooltip for the link that opens a list from the manage lists drawer."
+    },
+    "link_label_view_watchlist": {
+      "default": "View watchlist",
+      "description": "Aria-label for the link that opens the watchlist from the manage lists drawer."
+    },
+    "tooltip_view_watchlist": {
+      "default": "View watchlist",
+      "description": "Tooltip for the link that opens the watchlist from the manage lists drawer."
+    },
     "button_label_add_to_personal_list": {
       "default": "Add \"{title}\" to {list}",
       "description": "Aria-label for the button that allows users to add a movie or show to a personal list.",
--- a/projects/client/src/lib/components/buttons/watchlist/WatchlistButton.svelte
+++ b/projects/client/src/lib/components/buttons/watchlist/WatchlistButton.svelte
@@ -21,6 +21,7 @@
     type,
     onAdd,
     onRemove,
+    action,
     ...props
   }: WatchlistButtonProps = $props();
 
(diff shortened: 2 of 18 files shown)


ElMagnea (changes requested):
I think we're turning this into an Excel sheet. I'm thinking of different ways of solving this. I totally understand the need here, I just want it to look less visually heavy.

rectifyer:
@seferturan @ElMagnea Open to a different design, but I don't really like having to click another icon to enter and edit mode. That seems like too much functionality for this view?

PR #3167, opened 2026-08-22
Perf/reduce boot network cost

Marius-TV:
## Summary

Cuts network and parser work out of app boot. Driven by a HAR capture of
app.trakt.tv, in which **631 of 633 requests came from the service worker's
install-time precache** — not one from the page's own load path.

## What the HAR showed

Workbox precaches entries strictly one at a time
([workbox#2528](https://github.com/GoogleChrome/workbox/issues/2528)), and the
trace shows exactly that: ~200 ms per request, fully serialized, 2.49 MB over
139 s — and the capture was stopped *mid-install*, still working through JS
chunks. Globbing the real build output with the plugin's default pattern gives
the true scope: **884 entries / 22.8 MiB**, of which 845 are `_app/immutable/`
chunks and 2.9 MiB is `pwa/screenshots` + `yir` art.

Because precaching runs inside the install event's `waitUntil`, the worker
**could not activate for the entire duration**. For roughly the first three
minutes of a new visitor's session, no runtime cache served anything, while
~630 serialized requests competed with their navigation.

## Changes

Three focused commits:

1. **`perf(pwa)`: precache the app shell instead of the whole build.** Hashed
   build output doesn't need precaching — it's immutable by URL, so the
   `CacheFirst` static-asset route already in `service-worker.ts` caches exactly
   the chunks a visitor loads, when they load them. **884 entries / 22.8 MiB →
   13 entries / 61.5 KiB**, verified against a real build.

2. **`perf(player)`: load Plyr on demand.** `plyr.js` + `plyr.css` were fetched
   from cdn.plyr.io on every page for every visitor. `createPlyr` already owned
   the loading concern (it rAF-polled `globalThis.Plyr` for up to 10 s), so it
   now injects both on first trailer open, memoized, with retry on failure. Also
   retires the `media="print"` + `onload` swap that previously crashed
   `<svelte:head>` hydration replay (see `fix/plyr-css-onload-hydration`).

3. **`perf(pwa)`: inline the service worker registration.** `register-install.js`
   was an external `<script>` in `<head>` with **no `defer`** and a
   `?v=<timestamp>` query — a guaranteed-uncacheable round trip blocking HTML
   parsing on every page load. Inlining removes the request, and removes the
   staleness failure mode the timestamp was added for in 826447287 rather than
   working around it. That leaves `features/cache-bust/handle.ts` with no
   consumers, so it comes out of the hook sequence.

## Verification

- `deno task check`, `deno task lint`, `deno task test:unit` (**2879 passed**) —
  no new problems. Remaining check/lint output is pre-existing and in untouched
  files.
- Production-shaped build confirms the 13-entry manifest, and that neither
  `register-install` nor an eager `plyr.js` appears anywhere in the output.
- Each commit is independently self-consistent (verified: `register-install` is
  referenced *and* present through commit 2, and the cdn.plyr.io CSP directives
  are correctly retained for the on-demand loader).

## Reviewer notes

**Offline scope narrows, slightly.** Precaching everything nominally gave offline
access to any route. But `activate` purges the navigation cache on every update
and documents use `StaleWhileRevalidate`, so offline navigation *already* only
worked for previously-visited URLs — the 845 precached chunks had no document to
pair with. This trades a largely theoretical capability for a real one: the
worker now activates in seconds.

**Storage.** Immutable chunks now land in the `CacheFirst` static cache
(30-day expiration, no `maxEntries`) instead of the precache. In practice that's
less resident storage than the previous 22.8 MiB manifest, so I left it
uncapped — an LRU cap near one deploy's chunk count would thrash across
deploys. Flagging it rather than changing it.

**Not measured: LCP.** The capture began after page load, so it contains no
document, CSS, or font requests. Commits 2 and 3 are structurally sound (fewer
bytes, one less blocking round trip) but I have no before/after paint numbers —
worth a Lighthouse run to confirm.

**Next lever, not in this PR:** `nodes/0.js` is **885 KB raw / 306 KB gz** in
production and on the critical path for every page. It's the root layout, which
statically imports ~30 providers — several being dialogs and drawers that only
matter after an interaction. Splitting those needs a bundle-composition pass to
do safely.


seferturan (approved):
Applied some fixups:

- replaced the polling hack in `useInstallPrompt` with a `GlobalEventBus` listener, seeded from the stashed event. No more 2.5s window to miss the prompt.
  - typed `beforeinstallprompt` on `WindowEventMap` so the bus registration type-checks
- added specs for `useInstallPrompt` (stashed before mount + dispatched after mount) and `createPlyr` (retry after failed script/stylesheet)
- capped the static `CacheFirst` cache at 500 entries so it can't grow unbounded
- unified the Plyr failure paths into a single `settle()`, every path (timeout included) removes the tag
- typed `PlyrConstructor` instead of `...unknown[]`
- documented the imperative head injection pattern in `components.md` so feature-scoped assets stay off the boot path

PR #3174, opened 2026-08-24
fix(auth): collection of improvements

### LINKED ISSUE #3062: Authentication issues with magic link
### Code of Conduct

https://auth.trakt.tv/signin

I use the magic link login, as I created my account 10+ years ago with email/password and not with Apple/Google.  Since late April, my web session regularly signs itself out in the browser 1 week after the last browser login.  I am not a VIP user and only use my account on the first-party Trakt website and iOS app, not API keys, etc.

When the weekly browser logout happened this evening (Friday July 31 2026), I tried to log back in as usual.  I got the typical message about being emailed a magic link that will expire in 5 minutes but never got an email.  On any subsequent attempt to login, including from other devices, clicking the "Email me a signin link" button will give it a brief spinner then go back to the default button state as if I hadn't clicked it.  In the Developer Tools window (Edge), I see it reports the following error when this happens:
POST https://auth.trakt.tv/auth/magic 500 (Internal Server Error)

First and foremost, there seems to be an issue here with logging in at all for magic link, preventing access to the main site or forums.  Secondly, after that is addressed, it would be great to be able to stay logged in for more than a week at a time with a magic link if possible.

### Additional information

The iOS app has allowed me to stay logged in for a much longer time.  I am uncertain if I had been logged in there since before the switch to magic links.  However, this past Tuesday it hit an issue where the app seemed to consider me unauthenticated, necessitating reinstalling the app.  I logged back in with magic link in the iOS app on Tuesday and am still logged in there at this point, and nowhere else.  I can't speak to whether my iOS auth will have a 1 week limit until next Tuesday when I reach the 1 week mark since magic link login.

seferturan:
## 🎶 Notes 🎶

- Fixes #3062
- Guards the silent renew fired on window focus (cooldown + failure breaker)
  - A stale tab could self-sustain a renew loop: the renew iframe hands focus back to the top window, which refires the focus handler
  - Failures decay after 5 min, so a long lived tab doesn't get stuck after a few blips
- Only clears the session on a terminal OAuth error
  - Offline / timeout / 5xx / rate limit used to log you out too, even with a perfectly good refresh token still in storage
- Serializes renewal across tabs with a web lock
  - Refresh tokens rotate on use, so two tabs renewing at once means the second one presents a consumed token. Server reads that as a replay and invalidates the whole refresh family, on every device
  - `automaticSilentRenew` is off on purpose: it calls `signinSilent` straight off the manager, so it can't be routed through the lock. The expiry timer still comes from oidc-client-ts, we just own the subscription now
- Shows a snackbar when a login can't start
  - Was an uncaught rejection and a button that visibly did nothing
  - Only says "too many sign in attempts" when the status is actually readable. A rate limit at the edge strips the CORS headers off its own 429, so the browser reports a bare "Failed to fetch" with no status
- ⚠️ The weekly logout reports should be covered by the renewal changes, but that's inferred from the code, not reproduced. Will verify on prod

## 👀 Example 👀
Example when rate limited:
Screenshot 2026-08-24 at 14 23 28

### DIFF
--- a/projects/client/src/lib/features/auth/components/LoginErrorSnackbar.svelte
+++ b/projects/client/src/lib/features/auth/components/LoginErrorSnackbar.svelte
@@ -1,0 +1,35 @@
+<script lang="ts">
+  import Snackbar from "$lib/components/snackbar/Snackbar.svelte";
+  import * as m from "$lib/features/i18n/messages.ts";
+  import { loginErrorStore } from "../_internal/loginErrorStore.ts";
+  import { LoginErrorType } from "../models/LoginErrorType.ts";
+
+  const error$ = loginErrorStore.error$;
+
+  const content = $derived.by(() => {
+    switch ($error$) {
+      case LoginErrorType.RateLimited:
+        return {
+          title: m.text_login_rate_limited_title(),
+          message: m.text_login_rate_limited_message(),
+        };
+      case LoginErrorType.Unreachable:
+        return {
+          title: m.text_login_failed_title(),
+          message: m.text_login_failed_message(),
+        };
+      default:
+        return null;
+    }
+  });
+</script>
+
+{#if content}
+  <Snackbar
+    open
+    title={content.title}
+    message={content.message}
+    variant="error"
+    onDismiss={() => loginErrorStore.clear()}
+  />
+{/if}
--- a/projects/client/src/lib/features/auth/stores/initializeUserManager.ts
+++ b/projects/client/src/lib/features/auth/stores/initializeUserManager.ts
@@ -16,15 +16,13 @@ import { portWorkerAuthSession } from '../portWorkerAuthSession.ts';
 import { postToken } from '../postToken.ts';
 import { renewAccessToken } from '../renewAccessToken.ts';
 import { resolveOidcAuthority } from '../resolveOidcAuthority.ts';
+import { NOOP_FN } from '$lib/utils/constants.ts';
 import { safeLocalStorage } from '$lib/utils/storage/safeStorage.ts';
-import { time } from '$lib/utils/timing/time.ts';
+import { silentRenewPolicy } from '../silentRenewPolicy.ts';
 import { setToken, type Token } from '../token/index.ts';
 import type { AuthContextType } from './createAuthContext.ts';
+import { setSilentRenewGuard } from './silentRenewGuard.ts';
 import { setUserManager } from './userManager.ts';
-
-const renewCooldown = time.seconds(30);
-const maxRenewFailures = 3;
-const renewFailureReset = time.minutes(5);
 
 type InitializeUserManagerParams = {
   ctx: AuthContextType;
@@ -65,11 +63,11 @@ export function initializeUserManager(
       getOidcConfig(),
     );
 
-    const renewGuard = createSilentRenewGuard({
+    let retryHandle: ReturnType<typeof setTimeout> | null = null;
+
+    const renewGuard = createSilentRenewGuard<User | null>({
       now: () => Date.now(),
-      cooldownMs: renewCooldown,
-      maxConsecutiveFailures: maxRenewFailures,
-      failureResetMs: renewFailureReset,
+      ...silentRenewPolicy,
     });
 
     const syncToken = (user: User | null) => {
@@ -122,13 +120,24 @@ export function initializeUserManager(
       );
     };
 
+    const scheduleRenewRetry = () => {
+      if (retryHandle != null) {
+        return;
+      }
+
+      retryHandle = setTimeout(() => {
+        retryHandle = null;
+        renewIfLapsed();
+      }, silentRenewPolicy.cooldownMs);
+    };
+
     const handleSilentRenewFailure = async (error: unknown) => {
       if (isRateLimitError(error) && isInitializing.value) {
         dispatchRateLimitError();
       }
 
       if (isFatalRenewError(error)) {
-        await manager.removeUser().catch(() => {});
+        await manager.removeUser().catch(NOOP_FN);
         handleUserEvent(null);
         return;
       }
@@ -140,12 +149,20 @@ export function initializeUserManager(
       const current = await manager.getUser().catch(() => null);
       if (current?.refresh_token == null) {
         handleUserEvent(null);
+        return;
       }
+
+      scheduleRenewRetry();
     };
 
     const renewSilently = () => {
       renewGuard
         .renew(() => renewAccessToken(manager))
+        .then(({ outcome }) => {
+          if (outcome === 'blocked') {
+            scheduleRenewRetry();
+          }
+        })
         .catch(handleSilentRenewFailure)
         // The guard resolves without attempting once its breaker is open, and
         // a gated tree that never hears back renders nothing at all.
@@ -175,13 +192,31 @@ export function initializeUserManager(
       syncToken(user);
     };
 
-    const checkTokenOnFocus = async () => {
-      const user = await manager.getUser();
-      if (!user?.expired) {
+    // A tab that opens already focused never fires `focus`, so a failed
+    // renewal would sit lapsed until a manual reload.
+    function renewIfLapsed() {
+      manager.getUser()
+        .then((user) => {
+          if (!user?.expired) {
+            return;
+          }
+
+          renewSilently();
+        })
+        .catch(NOOP_FN);
+    }
+
+    const renewOnReconnect = () => {
+      renewGuard.reset();
+      renewIfLapsed();
+    };
+
+    const renewIfVisible = () => {
+      if (globalThis.document.visibilityState !== 'visible') {
         return;
       }
 
-      renewSilently();
+      renewIfLapsed();
     };
 
     const disposeExpiring = manager.events.addAccessTokenExpiring(
@@ -197,16 +232,30 @@ export function initializeUserManager(
 
     manager.getUser().then(initializeUser);
 
-    globalThis.window.addEventListener('focus', checkTokenOnFocus);
+    globalThis.window.addEventListener('focus', renewIfLapsed);
+    globalThis.window.addEventListener('online', renewOnReconnect);
+    globalThis.document.addEventListener('visibilitychange', renewIfVisible);
 
+    setSilentRenewGuard(renewGuard);
     setUserManager(manager);
 
     return () => {
+      if (retryHandle != null) {
+        clearTimeout(retryHandle);
+        retryHandle = null;
+      }
+
       disposeExpiring();
       disposeExpired();
       disposeUserLoaded();
       disposeUserUnloaded();
-      globalThis.window.removeEventListener('focus', checkTokenOnFocus);
+      globalThis.window.removeEventListener('focus', renewIfLapsed);
+      globalThis.window.removeEventListener('online', renewOnReconnect);
+      globalThis.document.removeEventListener(
+        'visibilitychange',
+        renewIfVisible,
+      );
+      setSilentRenewGuard(null);
       setUserManager(null);
     };
   });
--- a/projects/client/src/lib/features/auth/stores/initializeUserManager.ts
+++ b/projects/client/src/lib/features/auth/stores/initializeUserManager.ts
@@ -24,6 +24,16 @@ import { setUserManager } from './userManager.ts';
 const renewCooldown = time.seconds(30);
 const maxRenewFailures = 3;
 const renewFailureReset = time.minutes(5);
+const renewLockName = 'trakt-auth-renew';
+
+async function withRenewLock(task: () => Promise<void>) {
+  if (!navigator.locks) {
+    await task();
+    return;
+  }
+
+  await navigator.locks.request(renewLockName, task);
+}
 
 type InitializeUserManagerParams = {
   ctx: AuthContextType;
@@ -141,9 +151,28 @@ export function initializeUserManager(
       isInitializing.next(false);
     };
 
+    const renewUnderLock = async () => {
+      await withRenewLock(async () => {
+        const current = await manager.getUser();
+
+        const expiringAt = Date.now() + time.seconds(
+          manager.settings.accessTokenExpiringNotificationTimeInSeconds,
+        );
+        const didAnotherTabRenew = current != null && !current.expired &&
+          time.seconds(current.expires_at ?? 0) > expiringAt;
+
+        if (didAnotherTabRenew) {
+          handleUserEvent(current);
+          return;
+        }
+
+        await manager.signinSilent();
+      });
+    };
+
     const renewSilently = () => {
       renewGuard
-        .renew(() => manager.signinSilent())
+        .renew(renewUnderLock)
         .catch(handleSilentRenewFailure);
     };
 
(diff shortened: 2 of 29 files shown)


projects/client/src/lib/features/auth/stores/useAuth.ts
```diff
@@ -39,12 +42,19 @@ export function useAuth() {
...
+    loginErrorStore.clear();
+
+    try {
+      await manager?.signinRedirect({
+        extraQueryParams: {
+          hide_email_form: 'true',
+          lang,
+        },
+      });
+    } catch (error) {
```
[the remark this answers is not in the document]
seferturan: Changed it 👌 optional chaining silently swallowing the redirect is bad regardless.

Don't think the scenario is reachable though. The login buttons only render inside AuthProvider's gate, and `setUserManager` runs synchronously in its `onMount`, before anything flips `isInitializing` to false. So there's no clickable button while the manager is null.

`useRedirect` is the interesting one: it calls `login()` from a child `onMount`, which does run before AuthProvider's, so `getUserManager()` really is null there. But it's gated on `!isAuthorized.value`, and children only render that early when `hasSession` was true, which seeds `isAuthorized` true. Mutually exclusive.

Kept the guard anyway, that gate's ordering is fragile enough as is.

projects/client/src/lib/features/auth/stores/initializeUserManager.ts
```diff
@@ -107,15 +132,38 @@ export function initializeUserManager(
...
+        return;
+      }
+
+      isInitializing.next(false);
+    };
+
+    const renewUnderLock = async () => {
+      const current = await manager.getUser();
+
+      if (current && !current.expired) {
```
rudf0rd, line 150: `addAccessTokenExpiring` fires 60s before expiry by default, and `User.expired` is `expires_in <= 0`, so `current.expired` is false every time this handler runs.  looks like it re-posts the same token and returns without ever hitting `signinSilent`.  `getUser()` re-arms the timer for 1s, that re-fire lands inside the 30s cooldown, and focus only renews once it's already expired.  so a focused tab just runs out and the next 401 does `removeUser` + reload.  was the check meant to catch "another tab already renewed while i waited on the lock"?  if so maybe compare `expires_at` to what we had before taking the lock instead.

projects/client/src/lib/features/auth/stores/initializeUserManager.ts
```diff
@@ -142,19 +190,24 @@ export function initializeUserManager(
         return;
       }
 
-      manager.signinSilent().catch(handleSilentRenewFailure);
+      renewSilently();
     };
 
-    manager.getUser().then(initializeUser);
+    const disposeExpiring = manager.events.addAccessTokenExpiring(
```
rudf0rd, line 196: with `automaticSilentRenew` off we also lose the library's 5s retry on `ErrorTimeout`.  the expiring timer is one-shot per `events.load`, there's no `addAccessTokenExpired` listener, and `handleSilentRenewFailure` doesn't schedule anything on a transient error.  so a 10s iframe timeout on the expiring event means the token just lapses unless the window gets a focus event.  a retry timer or an expired handler with the guard shaping the cadence would cover it, but you may have already ruled that out.

projects/client/src/lib/features/auth/stores/initializeUserManager.ts
```diff
@@ -107,15 +132,38 @@ export function initializeUserManager(
...
         dispatchRateLimitError();
       }
-      handleUserEvent(null);
+
+      if (isFatalRenewError(error)) {
+        handleUserEvent(null);
+        return;
+      }
+
+      isInitializing.next(false);
```
rudf0rd, line 144: on a transient failure `isAuthorized` stays true and `getToken()` still hands out the expiring bearer.  once it lapses, the 401 in `createAuthenticatedFetch` runs `removeUser()`, which wipes the refresh token too.  main flipped to anonymous right away, this feels like it trades that for a hard logout with nothing left to renew from.  clearing `isAuthorized` without touching storage on the expired + transient case would leave the refresh token around for the next attempt.

projects/client/src/lib/features/auth/stores/initializeUserManager.ts
```diff
@@ -107,15 +132,38 @@ export function initializeUserManager(
     };
 
     const handleSilentRenewFailure = (error: unknown) => {
-      if (error instanceof Error && error.message.includes('429')) {
+      if (isRateLimitError(error)) {
```
rudf0rd, line 135: a 429 on a background renew is treated as transient below (session kept), but this still dispatches `FETCH_ERROR_EVENT`, and `ErrorProvider` swaps the whole page for `ErrorServicePage`.  so an idle tab with a perfectly good session gets a full-screen error until the next navigation.  worth limiting the dispatch to the init / callback paths where the user is actually stuck?

projects/client/src/lib/features/auth/stores/initializeUserManager.ts
seferturan: Yes, `addAccessTokenExpiring` fires ~60s early so `current.expired` was still false and the old guard returned without ever calling `signinSilent`. Reworked `renewUnderLock` to capture `expires_at` before taking the cross-tab lock, then only adopt a peer token when its expiry actually moved past that (didAnotherTabRenew); otherwise it renews.

projects/client/src/lib/features/auth/stores/initializeUserManager.ts
seferturan: Good catch. Since `automaticSilentRenew` is off, a proactive renew that timed out left nothing to try again at real expiry. Added an `addAccessTokenExpired` fallback listener that runs the same `renewSilently` path, so if the early renew fails we still get one more attempt at actual expiry. The `createSilentRenewGuard` cooldown keeps that from turning into a retry storm

projects/client/src/lib/features/auth/stores/initializeUserManager.ts
seferturan: Agreed, this was the nasty one. On a transient failure we now check the current user: if the token has actually expired we flip to anonymous via `handleUserEvent(null)` instead of holding a dead bearer. That avoids the 401 that would call `removeUser()` and clear the stored refresh token, so a later renew can still recover from storage. A still-valid token is left untouched

projects/client/src/lib/features/auth/stores/initializeUserManager.ts
seferturan: Right, a background renew shouldn't replace the whole page. The rate-limit dispatch is now gated on `isInitializing.value`, so the blocking `ErrorServicePage` only shows while the user is still stuck on initial auth. A 429 during a background renew keeps the existing session and stays silent.

projects/client/src/lib/features/auth/stores/initializeUserManager.ts
```diff
@@ -106,16 +131,53 @@ export function initializeUserManager(
...
+
+    const renewUnderLock = async () => {
+      const before = await manager.getUser();
+      const knownExpiresAt = before?.expires_at ?? 0;
+
+      await withRenewLock(async () => {
+        const current = await manager.getUser();
+
+        const didAnotherTabRenew = current != null && !current.expired &&
+          (current.expires_at ?? 0) > knownExpiresAt;
```
rectifyer: This check only catches a peer renew that lands between reading `before` and acquiring the lock. If this tab's expiring timer fires after another tab already renewed (background-tab timer throttling makes that a common ordering), `before` is already the fresh token, so `knownExpiresAt` is the new expiry, `didAnotherTabRenew` stays false, and we call `signinSilent()` on a token that is nowhere near expiring. The lock keeps it safe (no replay), but it rotates the refresh family and burns a round trip per throttled tab per expiry cycle.

Suggest making the in-lock check absolute instead of relative: adopt whenever `current` is not expired and its `expires_at` is beyond now plus the expiring-notification window (60s default). That covers both orderings and lets you drop the `before` read entirely.

projects/client/src/lib/features/auth/stores/initializeUserManager.ts
```diff
@@ -106,16 +131,53 @@ export function initializeUserManager(
...
+        }
+
+        await manager.signinSilent();
+      });
+    };
+
+    const renewSilently = () => {
+      renewGuard
+        .renew(renewUnderLock)
+        .catch(handleSilentRenewFailure);
```
rectifyer: `handleSilentRenewFailure` is async and awaits `manager.getUser()`. If that rejects (storage access can throw in degraded contexts), this `.catch` produces a new rejected promise that nobody handles, which is the same unhandled-rejection failure mode this PR removes from `login()`. A trailing `.catch(() => {})` here, or a try/catch around the `getUser()` inside the handler, closes it.

projects/client/src/lib/features/auth/stores/initializeUserManager.ts
```diff
@@ -142,19 +204,28 @@ export function initializeUserManager(
...
+    const disposeExpiring = manager.events.addAccessTokenExpiring(
+      renewSilently,
+    );
+    const disposeExpired = manager.events.addAccessTokenExpired(
+      renewSilently,
+    );
     manager.events.addUserLoaded(handleUserEvent);
     manager.events.addUserUnloaded(() => handleUserEvent(null));
-    manager.events.addSilentRenewError(() => handleUserEvent(null));
+    manager.events.addSilentRenewError(handleSilentRenewFailure);
```
rectifyer: With `automaticSilentRenew: false` the `SilentRenewService` never starts, so this event never fires; every failure now arrives through the guard's `.catch`. Either drop the listener or add a one-liner saying it is intentionally dormant, otherwise it implies a code path that does not exist.

projects/client/src/lib/features/auth/stores/initializeUserManager.ts
```diff
@@ -6,17 +6,35 @@ import { type User, UserManager } from 'oidc-client-ts';
...
 import { resolveOidcAuthority } from '../resolveOidcAuthority.ts';
 import { safeLocalStorage } from '$lib/utils/storage/safeStorage.ts';
+import { time } from '$lib/utils/timing/time.ts';
 import { setToken, type Token } from '../token/index.ts';
 import type { AuthContextType } from './createAuthContext.ts';
 import { setUserManager } from './userManager.ts';
 
+const renewCooldown = time.seconds(30);
+const maxRenewFailures = 3;
+const renewFailureReset = time.minutes(5);
```
rectifyer: After 3 transient failures with an expired token (laptop waking up offline is the easy repro), the tab flips to anonymous and the breaker blocks focus-triggered retries until this window lapses, even once the network is back. Consider funneling an `online` event into `renewSilently`, or resetting the breaker on the offline-to-online transition, so recovery does not wait up to 5 minutes. Fine as a follow-up since the refresh token survives in storage.

projects/client/src/lib/features/auth/stores/initializeUserManager.ts
```diff
@@ -106,16 +131,53 @@ export function initializeUserManager(
...
+      const current = await manager.getUser();
+      if (current?.expired ?? true) {
+        handleUserEvent(null);
+        return;
+      }
+
+      isInitializing.next(false);
+    };
+
+    const renewUnderLock = async () => {
```
rectifyer: The pure helpers are well tested, but the orchestration seam itself is not: the adopt-vs-renew decision here, the expired-vs-valid branch in `handleSilentRenewFailure`, and the no-Web-Locks fallback in `withRenewLock`. One spec driving a mocked `UserManager` through "peer renewed while this tab waited on the lock" and "transient failure with a still-valid token keeps the session" would cover exactly where the remaining risk lives.

projects/client/src/lib/features/auth/components/LoginErrorSnackbar.svelte
```diff
@@ -0,0 +1,44 @@
+<script lang="ts">
+  import Snackbar from "$lib/components/snackbar/Snackbar.svelte";
+  import * as m from "$lib/features/i18n/messages.ts";
+  import { onMount } from "svelte";
+  import { loginErrorStore } from "../_internal/loginErrorStore.ts";
+  import { LoginErrorType } from "../models/LoginErrorType.ts";
+
+  let error = $state<LoginErrorType | null>(null);
+
+  onMount(() => {
```
rectifyer: nit: RxJS observables satisfy Svelte's store contract, so this can shrink to `const error$ = loginErrorStore.error$;` and `$error$` in the template, dropping the `onMount`/`$state` bookkeeping.

rectifyer (dismissed):
Reviewed the latest head, and the earlier feedback rounds are all genuinely addressed (the `expires_at` comparison, the `addAccessTokenExpired` fallback, refresh-token preservation on transient expiry, and the `isInitializing` gate on the 429 dispatch). Requesting changes for two things in the renew path: the redundant refresh-token rotation in `renewUnderLock` when a tab's throttled timer fires late, and the unhandled rejection that `handleSilentRenewFailure` can itself produce. The rest of the comments are smaller (a dormant listener, breaker recovery after coming back online, a snackbar simplification, and a test-coverage gap on the orchestration seam). Details inline.

projects/client/src/lib/features/auth/stores/initializeUserManager.ts
seferturan: Good catch, the relative check only ever covered the "peer renewed while I waited on the lock" ordering. `getUser()` reads storage fresh every call, so a throttled tab that fires late already sees the peer's token as `before` and renews on top of it.

Went with your absolute check. Dropped the `before` read entirely and now adopt when the stored token is valid and still outside the expiring window. Pulled the window off `manager.settings.accessTokenExpiringNotificationTimeInSeconds` rather than hardcoding 60, so it cannot drift from the config. Compared in ms via `time.seconds(...)` since `expires_at` is epoch seconds 👌

projects/client/src/lib/features/auth/stores/initializeUserManager.ts
seferturan: Right. `safeLocalStorage` only wraps `setItem`, so `getItem` throws through, and `User.fromStorageString` can blow up on a corrupted entry too.

Went slightly further than the `.catch(() => {})` though. If that `getUser()` rejects, the bigger problem is `isInitializing` never flipping false, so the auth gate hangs on the spinner. It is now `await manager.getUser().catch(() => null)`, which falls into `current?.expired ?? true` and resolves to anonymous instead of stranding the gate.

projects/client/src/lib/features/auth/stores/initializeUserManager.ts
seferturan: Confirmed, `_raiseSilentRenewError` only has the two call sites inside `SilentRenewService._tokenExpiring`, and the service only starts from `startSilentRenew()` under `if (settings.automaticSilentRenew)`. We never call it manually, so the listener was dead.

Dropped it. Prefer that over a comment, we try to keep those out of the codebase.

projects/client/src/lib/features/auth/components/LoginErrorSnackbar.svelte
seferturan: Done 👌 `error$` is a `BehaviorSubject.asObservable()` so it emits synchronously, no undefined frame. We already do this in `ApiApplicationDetail.svelte` anyway, so this is the more consistent shape.

projects/client/src/lib/features/auth/stores/initializeUserManager.ts
seferturan: Agreed on the mechanism, though it takes 3 attempts spaced 30s apart to trip, so it is not instant. Leaving it for now as you suggest, a proper fix wants a `reset()` on the guard plus an `online` listener, which is new behaviour rather than a fix to what this PR changed. Refresh token survives in storage so nothing is lost, just slow. Will do it separately.

projects/client/src/lib/features/auth/stores/initializeUserManager.ts
seferturan: Fair, the adopt-vs-renew branch is exactly what I just rewrote and it has no coverage. Not in this push, will add the two cases you named on top. Skipping the `withRenewLock` fallback one, it is four lines and the mocking cost outweighs it.

Marius-TV (dismissed):

Marius-TV (dismissed):

PR #3180, opened 2026-08-25
fix(pwa): stop the service worker failing every intercepted request

seferturan:
## 🎶 Notes 🎶

- A Firefox user hit `NS_ERROR_INTERCEPTION_FAILED` on every `_app/immutable` chunk: the service worker intercepts each request but can never respond.
- Cause: `activate` awaited an unguarded `caches.delete`. On profiles where storage throws (ETP strict, clear-on-close, quota eviction) the rejected `waitUntil` fails activation, and because we `skipWaiting` + `claim`, the now-redundant worker still controls the page and fails every request. State-dependent, which is why it only shows on some profiles.
- Fix: the navigation cache is keyed to the build, so our `activate` cleanup is garbage collection rather than correctness. It no longer gates activation, and it is time-bounded so a hanging quota manager can't park the worker in `activating` either (the spec queues every fetch behind activation, same symptom with a timeout instead of an error).
  - Not airtight: workbox-precaching registers its own `activate` `waitUntil` and that one is unguarded, so the same storage profile can still fail activation there. Exposure is ~13 precache entries, and the client-side recovery below is the real backstop.
- Recovery for users already stuck: the chunk-failure listener never actually fired (resource errors don't bubble, needs capture). It now wipes caches + unregisters the worker client-side before reloading, instead of asking the broken worker to clean up via `_cb`.
  - Skipped entirely while offline, so a transient blip on mobile data doesn't cost the caches.
- Side fixes while in there:
  - `version.json` sat in CacheFirst for 30 days, killing SvelteKit's update polling.
  - static cache capped at 500 entries while one build emits ~850 files, so it evicted forever; content-hashed chunks get their own cache.
- ⚠️ Can't repro locally (needs a broken-storage profile), so partly a blind fix. The recovery path also covers unknown causes, will monitor after deploy.

### DIFF
--- a/projects/client/src/hooks.client.ts
+++ b/projects/client/src/hooks.client.ts
@@ -7,6 +7,8 @@ import { captureWebviewSession } from '$lib/features/webview/captureWebviewSessi
 import { SentryEndpoint } from '$lib/features/sentry/SentryEndpoint.ts';
 import { SENTRY_DSN } from '$lib/utils/constants.ts';
 import { safeSessionStorage } from '$lib/utils/storage/safeStorage.ts';
+import { delay } from '$lib/utils/timing/delay.ts';
+import { time } from '$lib/utils/timing/time.ts';
 import * as Sentry from '@sentry/sveltekit';
 import { handleErrorWithSentry } from '@sentry/sveltekit';
 import type { ErrorEvent as SentryErrorEvent } from '@sentry/sveltekit';
@@ -164,24 +166,33 @@ function getRejectionUrl(reason: unknown): string | undefined {
   return String((reason as { url?: unknown }).url ?? '');
 }
 
-function shouldReloadForRejection(event: PromiseRejectionEvent): boolean {
-  if (isDynamicImportError(event.reason)) return true;
-  return hasChunkPath(getRejectionUrl(event.reason));
+type RecoveryAction = 'purge-and-reload' | 'reload';
+
+function getRecoveryForRejection(
+  event: PromiseRejectionEvent,
+): RecoveryAction | undefined {
+  if (isDynamicImportError(event.reason)) return 'purge-and-reload';
+  if (hasChunkPath(getRejectionUrl(event.reason))) return 'purge-and-reload';
+  return undefined;
 }
 
-function shouldReloadForError(event: ErrorEvent): boolean {
-  if (isDynamicImportError(event.error ?? event.message)) return true;
-  if (event.target instanceof HTMLScriptElement) {
-    return hasChunkPath(event.target.src);
+function getRecoveryForError(event: ErrorEvent): RecoveryAction | undefined {
+  if (isDynamicImportError(event.error ?? event.message)) {
+    return 'purge-and-reload';
   }
-  return hasChunkPath(event.filename);
+
+  if (event.target instanceof HTMLScriptElement) {
+    return hasChunkPath(event.target.src) ? 'purge-and-reload' : undefined;
+  }
+
+  return hasChunkPath(event.filename) ? 'reload' : undefined;
 }
 
-function shouldReloadForClientEvent(
+function getRecoveryForClientEvent(
   event: ErrorEvent | PromiseRejectionEvent,
-): boolean {
-  if ('reason' in event) return shouldReloadForRejection(event);
-  return shouldReloadForError(event);
+): RecoveryAction | undefined {
+  if ('reason' in event) return getRecoveryForRejection(event);
+  return getRecoveryForError(event);
 }
 
 function buildReloadUrl() {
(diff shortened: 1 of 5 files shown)


projects/client/src/service-worker.ts
```diff
@@ -160,8 +179,7 @@ registerRoute(
     const hasCacheParam = url.searchParams.has('_cb');
 
     if (hasCacheParam) {
-      // Delete the entire navigation cache
-      await removeNavigationCache();
+      await Promise.all([removeNavigationCache(), removeAssetCaches()]);
```
rudf0rd: `_cb` is also the OAuth callback and VIP return param (`setCacheBuster` in `routes/callback/+page.svelte` and `useVip.ts`), so as far as I can tell this now wipes the immutable + static caches on every login and refetches all the chunks on the landing page. Before this PR it only dropped the navigation cache.

I also don't think the recovery flow reaches this — `hooks.client` unregisters the SW before `location.replace`, so that nav is uncontrolled. Either a separate param for the purge, or leaving the asset caches out of this route since the client already wipes them, would probably cover it.

projects/client/src/hooks.client.ts
```diff
@@ -155,27 +172,64 @@ function buildReloadUrl() {
...
+  const action = getRecoveryForClientEvent(event);
+  if (action === undefined) return;
+  triggerReloadOnce(action);
 }
 
 if (typeof window !== 'undefined') {
-  window.addEventListener('error', reloadOnceFromClientEvent);
+  // Resource load errors do not bubble.
+  window.addEventListener('error', reloadOnceFromClientEvent, {
+    capture: true,
```
rudf0rd, line 257: Is this too broad? Vite injects `modulepreload` links for every dep and treats preload failures as non-fatal, so a content blocker or a flaky preload for a route the user never visits would fire this and nuke every cache + the SW. Anyone with a persistent blocker would lose the whole cache every session. Could we limit it to `<script>` failures or the specific chunk the import actually needed?

projects/client/src/hooks.client.ts
```diff
@@ -129,24 +130,40 @@ function getRejectionUrl(reason: unknown): string | undefined {
...
+    return 'purge-and-reload';
   }
-  return hasChunkPath(event.filename);
+
+  const resourceUrl = getFailedResourceUrl(event.target);
+  if (resourceUrl !== undefined) {
+    return hasChunkPath(resourceUrl) ? 'purge-and-reload' : undefined;
+  }
+
+  return hasChunkPath(event.filename) ? 'reload' : undefined;
```
rudf0rd, line 184: Every prod frame lives under `/_app/immutable/`, so I think any uncaught exception in app code matches this. `'reload'` still goes through `_cb`, which now means the SW wipes the asset caches for a plain TypeError. Skipping `_cb` for the light reload, or using a different param so only the real recovery path purges, should sort it.

projects/client/src/hooks.client.ts
```diff
@@ -155,27 +172,64 @@ function buildReloadUrl() {
   return url.toString();
 }
 
-function triggerReloadOnce(): void {
+async function purgeCaches(): Promise<void> {
```
rudf0rd, line 200: Should this check `navigator.onLine` first? A PWA user on bad mobile data tapping a lazy route would hit "Failed to fetch dynamically imported module", and this deletes every cache on the origin (posters, fonts, precache) and unregisters the SW while they're offline. The dead end existed before, but losing everything cached for a transient blip looks new.

projects/client/src/service-worker.ts
seferturan: Woops, good catch 😅 Reverted `_cb` to navigation-cache-only and dropped `removeAssetCaches`. The recovery path already purges everything client-side before the reload, so the SW never needed to.

projects/client/src/hooks.client.ts
seferturan: Done 👌 Dropped the `<link>` check. A broken worker still surfaces as the dynamic import rejection right after, which we already catch, so preload failures stay non-fatal.

projects/client/src/hooks.client.ts
seferturan: Covered by reverting the `_cb` route above. A plain runtime error now only drops the navigation cache, same as before this PR.

projects/client/src/hooks.client.ts
seferturan: Done 👌 Recovery now bails when `navigator.onLine` is false, before consuming the once-per-session guard, so it retries once back online.

rectifyer (dismissed):

projects/client/src/service-worker.ts
```diff
@@ -92,12 +114,16 @@ self.addEventListener('install', () => {
...
+// `waitUntil` fails activation, leaving a redundant worker that still controls
+// its clients and fails every request it intercepts.
 self.addEventListener('activate', (event) => {
-  event.waitUntil(Promise.all([
-    removeNavigationCache(),
-    self.clients.claim(),
-  ]));
+  event.waitUntil(
+    Promise.all([
+      removeNavigationCache(),
```
rudf0rd: Now that the nav cache is sha-keyed, the new worker never reads from the old one, so this cleanup doesn't need to gate activation anymore. The `.catch` covers a reject, but if storage *hangs* (Firefox's quota manager does that on the same kind of profile) we sit in `activating` forever, and the spec queues every fetch behind activation — same symptom as the bug, just a timeout instead of `NS_ERROR_INTERCEPTION_FAILED`.

Could just `event.waitUntil(self.clients.claim())` and let `removeNavigationCache()` run best-effort, or race it against a few seconds if you'd rather keep it inside the event lifetime.

projects/client/src/service-worker.ts
```diff
@@ -92,12 +114,16 @@ self.addEventListener('install', () => {
 });
 
 // Claim clients and purge the navigation cache so stale locale-specific HTML
-// (e.g. a poisoned ru-RU document) is evicted on update.
+// (e.g. a poisoned ru-RU document) is evicted on update. A rejected
+// `waitUntil` fails activation, leaving a redundant worker that still controls
+// its clients and fails every request it intercepts.
```
rudf0rd: FYI, non-blocking: workbox-precaching registers its own `activate` `waitUntil` (`cache.keys()` + `cache.delete()` over the precache) and that one's unguarded, so the same storage profile could still fail activation there. Exposure is tiny with 13 precache entries, it just means the client-side recovery is the real backstop rather than this being airtight. Might be worth softening "the broken state can't form anymore" in the description.

rudf0rd (dismissed):
Neither of these are blockers.  Approving and if they're issue you want to tackle, a follow up PR is kosher.

projects/client/src/service-worker.ts
seferturan: Good catch, the hang case is the one I missed. Rejects were covered, time wasn't.

Kept it inside the event lifetime but bounded: `clients.claim()` is the only thing activation waits on now, and `removeNavigationCache()` races a 3s timeout. Fire-and-forget outside `waitUntil` risks the worker being killed mid-delete, and it's only GC, so the race felt like the better trade.

projects/client/src/service-worker.ts
seferturan: Yep, checked and you're right: `PrecacheController` adds its own `activate` listener and `workbox-core`'s `waitUntil` has no catch, so a reject there still fails activation.

Softened the description: our cleanup no longer gates activation, but that narrows the window rather than closing it, and the client-side recovery is the real backstop. Called out the ~13 entries too. Not much we can do about workbox's own listener short of not using `precacheAndRoute`, which feels like a bad trade for the exposure.

PR #3183, opened 2026-08-26
refactor(websocket): remove real-time invalidation feature

Marius-TV:
## What

Removes the `websocket` feature entirely — the whole `src/lib/features/websocket/` directory and its `<WSInvalidator />` mount in `+layout.svelte` — plus the now-unused `wss://` sources from the CSP `connect-src`.

The socket opened an AWS API Gateway connection and pushed cache-invalidation keys (`show:watchlist`, `movie:watchlist`, `episode:watched`, `movie:watched`, `account:settings`), which `WSInvalidator` debounced into `invalidateAll` calls so an open tab refreshed when the account changed on another device.

## Why

It had neither reconnect nor keepalive. The socket was created in exactly one place — the auth-token subscription in `WSInvalidator` — so nothing reopened it after a drop. API Gateway always drops it eventually: a 10-minute idle-connection timeout, and a 2-hour hard maximum connection duration. Since inbound messages only arrive when the account changes elsewhere, the idle timeout was the normal case, meaning the feature was reliably dead on long-lived tabs while `createSocket` logged `WS Error:` to the console.

## Behavioral impact

Cross-device invalidation no longer happens at all. In practice this codifies what was already true for any tab open more than 10 minutes, but tabs did get working real-time updates for their first few minutes and now will not. Every explicit user action still invalidates its own queries through `useInvalidator` as before — only the passive push channel is gone.

## Scope

The shared stores it consumed (`useInvalidator`, `useImportInProgress`, `useClearInProgress`) all have many other callers and are untouched. No `InvalidateAction` member became dead — `InvalidateAction.CheckIn` is still used by `useCheckIn` and `useStopNowPlaying`. No remaining reference to `websocket`, `WSInvalidator`, `wss://` or `execute-api` anywhere in `src`, `test` or `e2e`.

## Testing

- `deno task test:unit` — 528 files, 3026 tests passing (down 1 file / 3 tests, exactly the deleted `createConnection.spec.ts`).
- `deno task check` — 7402 files, 0 errors, 0 warnings.
- `deno task lint` — 339 problems, down from 342 on `main`; the delta is only the removed files. All remaining are pre-existing.


seferturan (approved):
Agreed on removing this 👌

Main reason for me isn't the reconnect gap, it's that we have no mechanism for local invalidations. There was no sender or device filter, so doing something in my own browser had the same key pushed back down my own connection and invalidated it again a second later.

Also added a small change to `refetchOnWindowFocus` on the settings query.

In the future we should look into adding this feature back, in a more robust way.

PR #3184, opened 2026-08-26
fix: escapes user submitted markdown and json-ld

seferturan:
## 🎶 Notes 🎶

- Comments, reviews and trivia render user markdown through `{@html}`. marked ships zero sanitization by design, so `<script>` and `javascript:` links went straight to the DOM.
- All markdown now goes through `createSafeMarked`. Raw html tokens are escaped to visible text, link and image hrefs must pass `hasSafeUrlProtocol`.
  - Escaped rather than stripped: on a block token `text` is the whole chunk, so stripping would render `<div>my entire review</div>` as nothing.
- The link and image renderers build their own attributes, so the href we validate is the exact string the browser resolves.
  - Without that, an entity encoded colon (`javascript&colon;alert(1)`) slips past the protocol check and the html parser decodes it back to `javascript:` in the attribute.
- eslint rule bans raw `marked` imports outside the factory, so the unsafe path is unreachable. Catches named and namespace imports.
- The json-ld block is assembled as a raw script tag and `JSON.stringify` doesn't escape `<`, so an api title containing `</script>` broke out of the element. Escaped at the sink rather than in each of the four builders.
- No dompurify: it needs a DOM and we ssr on workers, so it would mean pulling `isomorphic-dompurify` + jsdom into the worker bundle. Escaping at token level works the same on both sides, and we allow zero user html anyway, which is stricter than dompurify's allowlist.
- `escapeHtml` was living in the yir tooltip, moved to a shared util.

### DIFF
--- a/projects/client/eslint.config.js
+++ b/projects/client/eslint.config.js
@@ -43,4 +43,24 @@ export default ts.config(
       ],
     },
   },
+  {
+    files: ['src/**'],
+    ignores: ['src/lib/utils/markdown/**'],
+
+    rules: {
+      'no-restricted-imports': [
+        'error',
+        {
+          'paths': [
+            {
+              'name': 'marked',
+              'importNames': ['Marked', 'marked', 'parse', 'parseInline'],
+              'message':
+                'Raw marked output is not XSS safe. Use createSafeMarked from $lib/utils/markdown/createSafeMarked.ts instead.',
+            },
+          ],
+        },
+      ],
+    },
+  },
 );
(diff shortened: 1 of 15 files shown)


vladjerca (approved):
Approach is solid. Escaping at the token level instead of pulling dompurify + jsdom into the worker bundle is the right call for our SSR setup, and building the link/image attributes ourselves is what makes the entity encoded colon case actually hold.

Verified a few of the claims locally rather than taking them on trust:

- the eslint rule fires on both `import { Marked }` and `import * as marked`, and still lets type-only imports through, so the renderers in `comments/_internal/marked/` stay legal
- the `escaped` branch in the `text` renderer is not dead code. marked sets `escaped: true` on inline text inside a raw block and passes it through unescaped by default, so that override closes a real hole
- backslash escapes are not double escaped. `\<b\>`, `\&`, `\*` all render byte identical to stock marked

Pushed one cleanup on top: `hasSafeUrlProtocol` was walking the string with `Array.from().filter().join()` against a constant named `LAST_CONTROL_CHARACTER_CODE = 0x20`, but `0x20` is space, not a control character. Swapped it for a single regex replace with an accurate name. Behavior is identical.

Rebased on latest main, `deno task check` and the full unit suite are green 👌🏼

vladjerca (approved):

PR #3186, opened 2026-08-26
Streaming: handle new younify return params

rectifyer:
# Summary

Younify is updating its web-app communication interface as it comes out of beta, and the query params it appends to our return url are changing. This updates the `/callback/streaming` route to understand the new interface while remaining fully backwards compatible with the current one, so both app versions work during the transition.

# What changed

The release version of the younify app returns different params than the beta:

- The service param is renamed from `yc_serviceId` to `yc_service_id`.
- The `yc_status` vocabulary changes from `connected|success|cancelled|error` to `succeeded|canceled|failed|expired`.

The callback route now reads `yc_service_id` first and falls back to `yc_serviceId`, and `toConnection` accepts both status vocabularies: `succeeded` joins the connected statuses, `canceled` joins cancelled, and `failed` (plus any unknown value) maps to the existing error result. The old params can be dropped once the transition window closes.

The launch side already sends `return_url` in the connect request body, so the newly mandatory return url field required no changes.

# New expired status

The new `expired` status gets first-class handling instead of collapsing into the generic error banner, since it means the connect session timed out rather than something going wrong:

- Added `expired` to `StreamingConnectionStatusKind` and to the result kinds the settings page accepts.
- The snackbar shows dedicated copy ("Connection timed out" / "Your session expired. Try connecting {service} again.") with informational styling rather than the error variant.
- Two new i18n keys added to `i18n/meta/en.json`.

### DIFF
--- a/projects/client/i18n/meta/en.json
+++ b/projects/client/i18n/meta/en.json
@@ -9599,6 +9599,14 @@
       "default": "You can connect {service} whenever you're ready.",
       "description": "Message shown when the user cancelled connecting a streaming service."
     },
+    "text_connection_status_expired_title": {
+      "default": "Connection timed out",
+      "description": "Title shown in the banner when the streaming service connect session expired before completing."
+    },
+    "text_connection_status_expired_message": {
+      "default": "Your session expired. Try connecting {service} again.",
+      "description": "Message shown when the streaming service connect session expired before completing."
+    },
     "text_streaming_service_fallback": {
       "default": "your streaming service",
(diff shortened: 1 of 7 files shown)


seferturan (approved):
Applied some fixups:

- Status mapping is a lookup table now, instead of the if chain
- `toConnection` returns the union instead of a bare `string`, drift between the two ends is a compile error now rather than a banner that silently never shows
- `toKind` narrows off a typed array, dropped the two `as` casts

PR #3194, opened 2026-08-27
fix(spoiler): reveals blurred content on click everywhere

### LINKED ISSUE #3165: Inconsistent Mouse Interaction on Spoilers
### Code of Conduct

v3 Desktop with a mouse with the "show spoilers" setting disabled.

The cursor and onClick behavior of spoilers is inconsistent.
In my eyes, spoilers should always have a pointer cursor and always reveal the content on click, like it was in v2.

| Component | Cursor | onClick behavior |
| --- | --- | --- |
| Media Detail Page: Reviews Section | pointer | opens review drawer (content still blurred there) |
| Review Drawer: Review | pointer | reveals content |
| Media Card: Title | default | nothing
| Episode Drawer: Title | default | nothing
| Media Detail Page: Description | default | reveals content
| Summary Card (History): Title | pointer (only inherited from link) | opens card link

Could be I'm missing some other components

### Additional information

_No response_

seferturan:
## 🎶 Notes 🎶

- Fixes #3165
- Improves the spoiler reveal interaction by ensuring that clicking anywhere within a blurred spoiler container, including its child elements, will correctly reveal the content.

### DIFF
--- a/projects/client/src/lib/features/spoilers/components/Spoiler.svelte
+++ b/projects/client/src/lib/features/spoilers/components/Spoiler.svelte
@@ -17,6 +17,7 @@
 <trakt-spoiler
   use:spoiler={isSpoilerHidden}
   use:spoilMeAnyway={variant !== "persistent"}
+  data-variant={variant}
 >
   {@render children()}
 </trakt-spoiler>
@@ -40,9 +41,17 @@
       :global(p:not(button p):not(a p)),
       :global(.trakt-comment p),
       :global(span:not(button span):not(a span)) {
-        pointer-events: none;
-
         @include spoiler-blur();
+      }
+
+      :global(p:not(button p):not(a p)),
+      :global(.trakt-comment p),
+      :global(span:not(button span):not(a span)) {
+        pointer-events: none;
+      }
+
+      &[data-variant="dismissible"] {
+        cursor: pointer;
       }
     }
   }
--- a/projects/client/src/lib/features/spoilers/components/spoilMeAnyway.ts
+++ b/projects/client/src/lib/features/spoilers/components/spoilMeAnyway.ts
@@ -8,8 +8,18 @@ export function spoilMeAnyway(node: HTMLElement, isDismissible = true) {
       return;
     }
 
+    const { target } = e;
+    const isHidden = node.classList.contains(SPOILER_CLASS_NAME) ||
+      target.classList.contains(SPOILER_CLASS_NAME);
+
+    if (!isHidden) {
+      return;
+    }
+
+    e.preventDefault();
+
     node.classList.remove(SPOILER_CLASS_NAME);
-    e.target.classList.remove(SPOILER_CLASS_NAME);
+    target.classList.remove(SPOILER_CLASS_NAME);
   }
 
   node.addEventListener('click', handleClick);
--- a/projects/client/src/lib/features/spoilers/components/spoilMeAnyway.spec.ts
+++ b/projects/client/src/lib/features/spoilers/components/spoilMeAnyway.spec.ts
@@ -30,6 +30,31 @@ describe('action: spoilMeAnyway', () => {
     component.destroy();
   });
 
+  it('should prevent the default action when revealing', async () => {
+    const commentNode = document.createElement('div');
+    commentNode.classList.add(SPOILER_CLASS_NAME);
+
+    const component = await renderStore(() => spoilMeAnyway(commentNode, true));
+    const event = new MouseEvent('click', { bubbles: true, cancelable: true });
+    commentNode.dispatchEvent(event);
+
+    expect(event.defaultPrevented).toBe(true);
+
+    component.destroy();
+  });
+
+  it('should let the click through when nothing is hidden', async () => {
+    const commentNode = document.createElement('div');
+
+    const component = await renderStore(() => spoilMeAnyway(commentNode, true));
+    const event = new MouseEvent('click', { bubbles: true, cancelable: true });
+    commentNode.dispatchEvent(event);
+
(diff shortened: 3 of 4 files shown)


projects/client/src/lib/features/spoilers/components/spoilMeAnyway.ts
```diff
@@ -8,8 +8,18 @@ export function spoilMeAnyway(node: HTMLElement, isDismissible = true) {
...
+
+    if (!isHidden) {
+      return;
+    }
+
+    e.preventDefault();
+
     node.classList.remove(SPOILER_CLASS_NAME);
-    e.target.classList.remove(SPOILER_CLASS_NAME);
+    target.classList.remove(SPOILER_CLASS_NAME);
```
rectifyer, line 22: The hidden check only inspects `node` and `e.target`, but comment inline spoilers are `p.trakt-spoiler` where only the `span`s get `pointer-events: none` (`CommentBody.svelte`). A markdown link inside a blurred paragraph keeps pointer events, so the click target is the `<a>`: neither `node` nor `target` carries the class, the guard falls through, and the link navigates while the surrounding text is still blurred.

Walking up with `closest` covers descendants of the classed element; the `node.contains` guard keeps it from matching a wrapper above this action's own node (which has its own handler). The removal at the bottom then needs to target the matched element instead of the raw click target:

```suggestion
    const { target } = e;
    const hiddenTarget = target.closest(`.${SPOILER_CLASS_NAME}`);
    const isHidden = node.classList.contains(SPOILER_CLASS_NAME) ||
      (hiddenTarget !== null && node.contains(hiddenTarget));

    if (!isHidden) {
      return;
    }

    e.preventDefault();

    node.classList.remove(SPOILER_CLASS_NAME);
    hiddenTarget?.classList.remove(SPOILER_CLASS_NAME);
```

A spec dispatching the click from a descendant of the classed element (e.g. an `<em>` inside a `.trakt-spoiler` paragraph) would lock this in.

projects/client/src/lib/features/spoilers/components/Spoiler.svelte
```diff
@@ -37,12 +38,20 @@
...
       :global(p:not(button p):not(a p)),
       :global(.trakt-comment p),
       :global(span:not(button span):not(a span)) {
         pointer-events: none;
+      }
 
-        @include spoiler-blur();
+      &[data-variant="dismissible"] {
+        cursor: pointer;
       }
```
rectifyer, line 55: This only applies when the host itself is hidden (`trakt-spoiler.trakt-spoiler[data-variant="dismissible"]`), but `spoilMeAnyway` also reveals two surfaces that keep `cursor: default`:

- a flagged spoiler comment whose media spoiler isn't hidden: `.trakt-comment.trakt-spoiler` is blurred by `CommentBody.svelte`'s own SCSS and has `use:spoilMeAnyway` on the div
- inline `p.trakt-spoiler` spoilers inside comments

#3165 explicitly asks for a consistent pointer cursor, so those paths still reproduce the inconsistency. Suggest matching rules in `CommentBody.svelte`'s style block (scoped there since the preview variant is persistent and shouldn't advertise clickability):

```scss
.trakt-comment {
  &:global(.trakt-spoiler),
  :global(p.trakt-spoiler) {
    cursor: pointer;
  }
}
```

rectifyer (changes requested):
The core fix is correct: dropping `pointer-events: none` from the text-only host plus the `preventDefault()` reveal matches the intent of #3165, and the new tests cover both branches. Requesting changes for two remaining slices of the inconsistency the issue reports (details inline):

1. Clicks on children of an inline spoiler (e.g. a markdown link inside `p.trakt-spoiler`) slip through the hidden check and navigate while the text is still blurred.
2. The new pointer cursor only covers the `trakt-spoiler` host; blurred comments and inline comment spoilers still show a default cursor.

Non-blocking notes for follow-up:
- `preventDefault()` cancels native/router navigation but not ancestor JS click handlers; worth deciding whether a reveal click should also `stopPropagation()` before a dismissible spoiler ever sits inside an `onclick` container.
- Spoilers are now visibly interactive but not keyboard-operable (no `role`/`tabindex`/Enter handling).
- The `p`/`.trakt-comment p`/`span` selector list now appears three times in `Spoiler.svelte`; an SCSS variable would keep them from drifting apart.

projects/client/src/lib/features/spoilers/components/spoilMeAnyway.ts
seferturan: No, `pointer-events` is inherited, so this doesn't reproduce.

A link *inside* the spoiler span picks up `pointer-events: none` from `p.trakt-spoiler span` (nothing re-enables it, `default-link-style` doesn't touch it), so the event retargets to the `p` and `target.classList.contains` already catches it.

A link *outside* the span isn't blurred at all. `CommentBody.svelte` blurs `p.trakt-spoiler span`, not the `p` itself, so navigating there is the correct behaviour. With `closest` we'd walk up to the `p`, `preventDefault()`, and break navigation on a fully visible link.

For a blurred ancestor to sit between `node` and `target` you'd need one whose descendants keep pointer events. The host retargets to itself, `.trakt-comment.trakt-spoiler` *is* the node, and `p.trakt-spoiler` isn't blurred itself. So no case left. Keeping as is.

projects/client/src/lib/features/spoilers/components/Spoiler.svelte
seferturan: Good catch, both of those still showed `default`. Added the rule to `CommentBody.svelte`'s style block 👌

PR #3211, opened 2026-09-04
feat(profile): split the up to date progress filter into up to date and ended

### LINKED ISSUE #3210: Split the "Up to date" progress filter into up to date and ended
### Code of Conduct

The `Up to date` filter in Profile > Progress answers two different questions at once.

It lists every show whose aired episodes I have all watched. So a series that wrapped up years ago sits right next to a series that is still airing and simply has nothing unwatched at the moment. Those are very different things: one is finished and can be filed away, the other is only waiting for next week's episode.

From the progress list there is no way to tell them apart without opening each show.

Split `Up to date` into two filters:

- **Up to date** - still airing, and I am caught up.
- **Ended** - the show is over (status `ended` or `canceled`).

`In progress` and `Dropped` stay as they are.

This needs no API change. `up_next_nitro` already returns the full show object, `status` included, even though the contract documents `status` as only available with `extended=full`. Verified against the live API with my own account:

```
GET /sync/progress/up_next_nitro?intent=completed&limit=100

total completed: 16
  ended              10
  returning series    6
status missing:       0
```

The show keys that come back:

```
aired_episodes, airs, available_translations, certification, colors,
comment_count, country, first_aired, genres, homepage, ids, images,
language, languages, last_aired, network, original_title, overview,
rating, runtime, social_ids, status, subgenres, tagline, title,
total_runtime, trailer, updated_at, votes, year
```

The client already carries this through: `mapToShowEntry` maps `status: show.status ?? 'unknown'`, and `mapToWatchedProgressEntry` runs every up next entry through it. So the value is on `ProgressEntry.show.status` today, unused.

That means both filters can share one query and one request, partitioning the response on the status. Switching between `Up to date` and `Ended` costs no extra fetch.

An unknown status should count as still airing, so behaviour only changes for shows Trakt is explicit about.

**A status filter on the endpoint.** Cleaner server side, but the split does not need it, and it would be a bigger ask for a narrower payoff.

**Leaving it to sorting.** Sorting by status is not the same thing; it still mixes both groups in one list and does not let you look at just the finished ones.

### Additional context

I have this working on my fork and can open the PR right away: 6 files, one new util, plus specs covering the partition. Screenshots in the PR.

Related: I raised the same idea for the Android app in trakt/trakt-android#353, and I am following the guidance to bring it to web first.

### What it looks like

Implemented in #3211 and running against my own account, so these are live API responses.

**`Up to date` after the split - 6 shows, all still airing**

**`Ended` - the 10 shows that are actually over**

Together they are exactly the 16 that the single `Up to date` filter listed before, and each one matches the status the API returns for it.

### Prior art on the API side

This is not a new ask. trakt/trakt-api#187, *"Making it easier to filter out completed shows"*, ran from 2020 to April 2026 with several third-party app developers asking for the same distinction, because a show being finished and a show merely being caught up are not the same state to a user. One of them hit exactly the trap that makes this a server-informed question rather than something a client can cache:

> I thought for a second I might cache locally if a show has ended, but trouble is some shows un-end, like Will & Grace or Gilmore Girls.

That issue was closed once the up next endpoint grew `intent`, which is what makes the buckets addressable in the first place. The status that separates them already rides along in the same response - this request is just about surfacing that split in the UI.

matheus-souza:
Closes #3210

### Why

The `Up to date` filter in Profile > Progress answers two questions at once. It lists every show whose aired episodes are all watched, so a series that wrapped up years ago sits next to a series that is still airing and simply has nothing unwatched right now. From the progress list there is no way to tell which shows are actually finished without opening each one.

### What changed

`Up to date` becomes two filters: **Up to date** (still airing, caught up) and **Ended** (status `ended` or `canceled`). `In progress` and `Dropped` are untouched.

No API change is involved. `up_next_nitro` already returns the full show object with `status`, and `mapToShowEntry` already maps it onto `ProgressEntry.show.status` - the value was simply unused. Verified against the live API:

```
GET /sync/progress/up_next_nitro?intent=completed&limit=100
  16 shows -> 10 ended, 6 returning series, 0 without a status
```

- **Both filters share one query.** `typeToQuery` maps `completed` and `ended` to the same `progressWatchedQuery({ intent: 'completed' })`, so the dependencies match and toggling between them costs no request. The split happens in the rxjs pipe, before the intl overlay, so discarded entries never get an overlay fetch.
- **`hasEnded` lands in `lib/utils/media/`** next to `hasAired`, since it is a question about a show rather than about this screen. An unknown status counts as still airing, which preserves today's behaviour.
- **`ProgressListType` is exported from `useProgressList.ts`** and reused by `ProgressPaginatedList` and `ProgressItem`. The union was previously written out three times and would have drifted.
- **The progress CTA condition became `!== 'dropped'`.** It was `in-progress || completed`, which is the same set today, but as written it would have silently left `Ended` with no CTA and no empty-state copy.

### Localisation

Two new keys in `i18n/meta/en.json`, `button_text_progress_ended` and `button_label_progress_ended`, following the existing `button_*_progress_*` pattern with descriptions. They ship English-only and pick up translations on the next Crowdin sync.

I deliberately did not touch `button_label_progress_completed` ("View completed shows"). With the split its wording is arguably stale, but changing it would invalidate that string across every locale for a small gain. Happy to change it if you would rather.

### Testing

- `hasEnded.spec.ts` - covers ended, canceled, the nine airing statuses, and unknown.
- `useProgressList.spec.ts` - serves a completed bucket holding one ended and one airing show through MSW, then asserts `Up to date` keeps only the airing one, `Ended` keeps only the finished one, and neither `In progress` nor `Dropped` filters by status. The dropped case is served an ended show on purpose, so the predicate leaking onto that branch would make the show vanish. Confirmed the split assertions fail when the predicate is inverted.

Both new files land at 100% line coverage.

```
deno task check    0 errors, 0 warnings
eslint             clean on all changed files
vitest             8 passed (2 files)
```

Full suite: 3134 passed, 1 unrelated timeout in `hooks.client.spec.ts` (a filesystem walk that passes in isolation on my machine).

### An empty segment, caught by running it

The progress toggler renders in its icon variant, and `ToggleIcon` resolves the glyph from the option value, so the new filter first shipped as a blank segment - the option was there and clickable, just invisible. Second commit maps `ended` to `FinaleIcon`, the checkered flag already used for season finales. The episode type toggler that owns it never shares a surface with this one.

### Screenshots

Captured against my own account on `dev:contrib`, so these are live API responses, not fixtures.

**`Up to date` - 6 shows, all still airing**

01-progress-up-to-date

**`Ended` - 10 shows, all finished**

02-progress-ended

**`In progress` - 14 shows, unchanged by this PR**

03-progress-in-progress

**`Dropped` - unchanged by this PR**

04-progress-dropped

Every show rendered under each filter was cross-checked against the status the API returned for it:

```
UI "up to date" (6)                        UI "ended" (10)
  paradise-2025      returning series        grace-and-frankie   ended
  georgie-mandy-...  returning series        i-will-find-you     ended
  the-rookie-2018    returning series        better-call-saul    ended
  the-pitt           returning series        the-big-bang-theory ended
  old-dog-new-tricks returning series        narcos              ended
  untamed            returning series        sherlock            ended
                                             hannibal            ended
                                             sons-of-anarchy     ended
                                             breaking-bad        ended
                                             house               ended
```

16 of 16 correct, and 6 + 10 is exactly the 16 the single `Up to date` filter listed before - nothing is dropped by the partition.

### DIFF
--- a/projects/client/i18n/meta/en.json
+++ b/projects/client/i18n/meta/en.json
@@ -8724,6 +8724,10 @@
       "default": "Up to date",
       "description": "Text for the button that allows users to filter the list to only display shows that are completed. This title should be as short and concise as possible."
     },
+    "button_text_progress_ended": {
+      "default": "Ended",
+      "description": "Text for the button that allows users to filter the list to only display shows that have ended or been canceled. This title should be as short and concise as possible."
+    },
     "button_text_progress_dropped": {
       "default": "Dropped",
       "description": "Text for the button that allows users to filter the list to only display shows that are dropped. This title should be as short and concise as possible."
@@ -8735,6 +8739,10 @@
     "button_label_progress_completed": {
       "default": "View completed shows",
       "description": "Aria-label for the button that allows users to filter the list to only display shows that are completed."
+    },
+    "button_label_progress_ended": {
+      "default": "View ended shows",
+      "description": "Aria-label for the button that allows users to filter the list to only display shows that have ended or been canceled."
     },
     "button_label_progress_dropped": {
       "default": "View dropped shows",
--- a/projects/client/src/lib/components/toggles/ToggleIcon.svelte
+++ b/projects/client/src/lib/components/toggles/ToggleIcon.svelte
@@ -91,6 +91,10 @@
   <TrackIcon />
 {/if}
 
+{#if option.value === "ended"}
+  <FinaleIcon />
+{/if}
+
 {#if option.value === "dropped"}
   <DropIcon />
 {/if}
--- a/projects/client/src/lib/components/toggles/_internal/constants.ts
+++ b/projects/client/src/lib/components/toggles/_internal/constants.ts
@@ -18,7 +18,7 @@ type DiscoverToggleType = DiscoverMode;
 type SocialToggleType = 'following' | 'followers' | 'requests';
 type CommentToggleType = CommentSortType;
 type TriviaToggleType = 'spoilers' | 'no-spoilers';
-type ProgressToggleType = 'in-progress' | 'dropped' | 'completed';
+type ProgressToggleType = 'in-progress' | 'dropped' | 'completed' | 'ended';
 type ActivityToggleType = 'reviews' | 'ratings';
 type LibraryToggleType = 'plex' | 'other';
 type EpisodeTypeToggleType = EpisodeTypeFilter;
@@ -128,6 +128,11 @@ const progress: ToggleDefinition<'progress'> = {
       value: 'completed',
       text: m.button_text_progress_completed,
       label: m.button_label_progress_completed,
+    },
+    {
+      value: 'ended',
+      text: m.button_text_progress_ended,
+      label: m.button_label_progress_ended,
     },
     {
       value: 'in-progress',
--- a/projects/client/src/lib/sections/profile/components/ProgressList.svelte
+++ b/projects/client/src/lib/sections/profile/components/ProgressList.svelte
@@ -16,7 +16,7 @@
   const { current, set, options } = useToggler("progress");
 
   const cta = $derived(
-    $current.value === "in-progress" || $current.value === "completed"
+    $current.value !== "dropped"
       ? {
           type: "progress" as const,
           mediaType: "show" as const,
--- a/projects/client/src/lib/sections/profile/components/ProgressPaginatedList.svelte
+++ b/projects/client/src/lib/sections/profile/components/ProgressPaginatedList.svelte
@@ -4,12 +4,13 @@
   import type { ListSortProps } from "$lib/sections/lists/user/models/ListSortProps.ts";
   import { useSort } from "$lib/sections/lists/user/useSort.ts";
   import ProgressItem from "./_internal/progress/ProgressItem.svelte";
-  import { useProgressList } from "./_internal/useProgressList.ts";
-
-  type ProgressType = "in-progress" | "completed" | "dropped";
+  import {
+    type ProgressListType,
+    useProgressList,
+  } from "./_internal/useProgressList.ts";
 
   type ProgressPaginatedListProps = {
-    type: ProgressType;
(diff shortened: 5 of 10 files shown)


matheus-souza:
Filed the API-side report for the counts I left out of this PR: trakt/trakt-api#926.

Worth flagging one consequence for this repo specifically, separate from the filter split. `extractPageMeta` reads `x-pagination-page-count`, and since up next always sends one, `defineQuery`'s `getNextPageParam` takes the `paginated` branch and computes `page.current < page.total` - which is `1 < 14` at `limit=100`. So the progress list will keep asking for pages that come back empty.

Not touching it here, since it is a different concern from the split and the real fix is server side. Happy to open it separately if you would rather have it tracked on this repo too.

matheus-souza:
The `DeepSource: Test coverage` failure here is `Timed out: Artifact was never reported`, not a coverage regression.

The workflow runs on this PR are all sitting at `action_required`, so `ci_cd.yml` never ran and the `coverage` job never uploaded `lcov.info`. Approving the workflow runs should produce the artifact. `DeepSource: JavaScript` passed, since that analysis does not depend on CI.

One thing I could not verify from previous fork PRs: the upload step uses `secrets.DEEPSOURCE_DSN`, which GitHub withholds from fork-triggered runs, and the step is `continue-on-error`. So the check may stay red even after approval. The last fork PRs that ran the full suite (#2924, #2933) predate this check being reported, so they do not settle it either way.

Locally the suite is green:

```
vitest        7 passed (2 files)   hasEnded.spec.ts, useProgressList.spec.ts
deno task check   0 errors, 0 warnings
eslint            clean on every changed file
```

Full run is 3134 passed with one unrelated timeout in `hooks.client.spec.ts`, a filesystem walk that passes in isolation on my machine.

Marius-TV (approved):

PR #3221, opened 2026-09-05
feat(toggle): let the progress filters be reordered by holding a chip

### LINKED ISSUE #3220: Let the progress filters be reordered by holding a chip
### Code of Conduct

The progress filters render in a fixed order - `Up to date`, `In progress`, `Dropped` - for everyone.

That order suits someone catching up on finished shows. It does not suit someone who opens Trakt to see what is still running: `In progress` is the only filter they ever want, and it always sits behind one they rarely consult. Every visit to Profile > Progress, and to the full progress page, costs an extra click to get past a filter that is not theirs.

`useToggler` already persists the *selected* filter in local storage, so the toggler is willing to remember a preference here. The order simply is not one of them.

Let the chips be dragged into whatever order suits the person using them.

1. **Holding a segment claims it** and slides straight into a drag - one continuous gesture, and the same one the Android app uses, so the two platforms feel alike. It works with a mouse and with touch.
2. **The row rocks gently while a chip is held**, so it reads as editable rather than broken, and the slot the chip would land in dims to show where it is going.
3. **The order persists** next to the selection, in the same store, and applies to both places the row appears: the Progress section and the full progress page.

Selecting a filter keeps doing exactly what it does today. The order is presentation only, so nothing is refetched when it changes.

Two details worth stating:

- Reordering should be **opt in per toggler**. Only progress asks for it; social, discover, activity and the rest keep their fixed order, since their order is not a personal preference in the same way.
- A stored order should be **reconciled against the current options** on read, so an order saved before a filter was added or removed still applies instead of being thrown away.

- **A "move first" item in a context menu on each chip.** Discoverable, but it needs a menu on a small pill and only expresses one of the moves.
- **A setting in Settings.** Further from the thing being ordered, and a preference nobody would go looking for.
- **Making `In progress` the default for everyone.** Trades one fixed order for another; whoever preferred `Up to date` first then pays the click.
- **A dedicated reorder drawer, like user lists have.** Right for a long list with per-item actions, far too heavy for three chips.

### Additional context

Same request for the Android app in trakt/trakt-android#363, where the filters have the same fixed order.

I have this implemented and verified in a browser, and will open the PR alongside this issue.

matheus-souza:
Closes #3220

### Why

The progress filters render in a fixed order, so `In progress` always sits behind `Up to date` even for someone who only ever consults it. `useToggler` already persists the selected filter, so the toggler was willing to remember a preference here - the order simply was not one of them.

### What changed

**`holdToReorder`**, a new action on the segment row. A press starts a hold; moving more than 10px before it lands reads as a scroll and cancels, so a tap and a swipe both still behave. Once the hold lands the gesture owns the pointer, `pointermove` picks the slot, and the release commits. It works with a mouse and with touch, and matches the gesture the Android app uses.

**`toReorderIndex`** and **`moveItem`** are the pure halves, kept out of the DOM so they can be tested - which slot the pointer is over, and the resulting order. `moveItem` clamps both ends, so a drag that finishes outside the row parks the chip at the nearest end rather than dropping it.

**`useToggler`** persists the order beside the selection, in the same store and the same shape. A stored order is reconciled against the current options on read: values that no longer exist are dropped and new ones appended, so an order saved before a filter was added or removed still applies.

**Reordering is opt in.** `SegmentedSelect` takes `reorderable` and `onReorder`, and only the progress toggler passes them. Every other toggler - social, discover, activity, library - is untouched and keeps its fixed order.

While a chip is held the row rocks and the target slot dims, so it reads as editable. Both are dropped under `prefers-reduced-motion`.

### Testing

Verified in a browser against my own account, on Profile > Progress:

```
initial      completed → in-progress → dropped
after drag   in-progress → dropped → completed
persisted    ["in-progress","dropped","completed"]
after reload in-progress → dropped → completed
```

The drag was a real hold-and-move: press, wait past the hold threshold, move to the far end, release. The other togglers on the same page (media/shows/movies, people/lists) did not move, which is the check that `reorderable` is actually scoped.

```
vitest             21 passed (3 files, incl. the existing useToggler spec)
deno task check    0 errors, 0 warnings
eslint             clean on every changed file
```

`moveItem.spec.ts` covers forward, backward, no-op, both clamped ends, never dropping an entry and an empty row. `toReorderIndex.spec.ts` covers the slot under the pointer, switching at the midpoint rather than the edge, and parking past either end.

### Screenshots

**Before: the default order**

01-before

**A chip held, the row in reorder mode**

02-holding

**After: `In progress` first, and it survives a reload**

03-after

### Worth a second opinion

This is the one of these where the interaction is a judgement call rather than a mechanical port. Hold-to-drag is a touch idiom first; on desktop it is less conventional than a drag handle. I went with it so the two platforms feel the same, and because a handle on a small pill is awkward - but if you would rather have a different affordance on web, the pure parts and the persistence stay as they are and only the action changes.

Keyboard reordering is not included. It would need an interaction of its own rather than falling out of this one, and I did not want to guess at it here.

### DIFF
--- a/projects/client/src/lib/components/select/SegmentedSelect.svelte
+++ b/projects/client/src/lib/components/select/SegmentedSelect.svelte
@@ -6,6 +6,8 @@
   import { appendGlobalParameters } from "$lib/features/parameters/appendGlobalParameters.ts";
   import type { SegmentedSelectOption } from "./models/SegmentedSelectOption.ts";
   import type { SegmentedSelectProps } from "./models/SegmentedSelectProps.ts";
+  import { moveItem } from "$lib/utils/array/moveItem.ts";
+  import { holdToReorder } from "./_internal/holdToReorder.ts";
   import {
     trackSelector,
     type TrackSelectorParams,
@@ -24,8 +26,14 @@
     collapsedCount = 0,
     expanded = false,
     extension,
+    reorderable = false,
+    onReorder,
     onChange,
   }: SegmentedSelectProps<TValue> = $props();
+
+  let heldIndex = $state<number | null>(null);
+  let previewIndex = $state<number | null>(null);
+  const isReordering = $derived(heldIndex !== null);
 
   const selectedIndex = $derived(
     Math.max(
@@ -53,7 +61,19 @@
     collapsed: boolean;
   };
 
+  const reorderTo = (from: number, to: number) => {
+    if (from === to) return;
+    onReorder?.(
+      moveItem({ items: options.map((option) => option.value), from, to }),
+    );
+  };
+
   const segmentProps = ({ option, index, collapsed }: SegmentPropsParams) => ({
+    "data-segment-index": index,
+    "data-held": heldIndex === index ? "true" : undefined,
+    "data-preview": previewIndex === index && heldIndex !== index
+      ? "true"
+      : undefined,
     class: [
       "segment",
       !collapsed && index === selectedIndex && "is-selected",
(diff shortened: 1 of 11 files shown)


seferturan:
Thanks for taking a crack at this :D Closing this one for now though:

- The progress toggle only has 4 options and they're all visible at once, so reordering can't really make anything easier to reach.
  - The selection is already remembered, so you land back on your filter anyway.
  - Beyond that, not a fan of freely re-orderable UI elements. Though maybe something to revisit when later on we have dedication settings/edits per list.

PR #3227, opened 2026-09-08
feat(errors): every failure gets a scene of its own

ElMagnea:
## Visuals

Before
Screenshot 2026-09-07 at 14 53 34

After
Screenshot 2026-09-08 at 13 00 37
Screenshot 2026-09-08 at 13 00 30
Screenshot 2026-09-08 at 13 00 22
Screenshot 2026-09-08 at 12 59 53

## What

Rebuilds the full-page error screens as one composed card, and gives each
failure a mark of its own instead of the Trakt double-check standing in for
all of them.

Every error now shares the same shell: an inset surface, a ringed mark on a
purple bloom, a scene kicker, the headline, the copy, and a stacked actions
column.

| Failure | Mark | Kicker |
| --- | --- | --- |
| Service unreachable / rate limited | a transmitter whose waves thin out to dots and pulse without ever landing | `INTERMISSION` |
| 404 | a film strip with its middle frame punched out, breathing | `MISSING REEL` |
| Unexpected error | a clapperboard that snaps shut and springs back open | `BAD TAKE` |
| Account locked | a projector aperture shut down to a keyhole (deliberately still) | `HOUSE CLOSED` |

Every animation sits behind `prefers-reduced-motion`, and all the card's
colours are new tokens defined in both theme mixins.

## How to review

`/_design_system/errors` flips through all four scenes. The marks are
animated, so give each one a few seconds - a still frame does not show what
they do. The active scene rides in `?scene=`, and the unexpected-error scene
carries a synthetic session id and stack so the details block is real.

The 404 is also on any unknown URL if you want it without the frame around it.

## Copy changes

- **Service unreachable** now reads "We lost the signal" over "Trakt can't be
  reached right now. Your watch history is safe." / "We'll pick up exactly
  where you left off." - a promise about the user's data, rather than "The
  connection is... absent. Like a forgotten memory."
- **The 404 aside loses the parentheses wrapping it**, in the English source.
  CrowdIn owns the translated catalogs, so the other locales pick this up
  through the sync rather than by hand.
- Four new kicker keys, and the service message splits across two keys so the
  reassurance is its own row.

Every catalog change flows from `meta/en.json` only - this PR touches no file
under `i18n/messages/`. Other locales keep their existing translations until
the CrowdIn sync carries these across.

## Button

Four opt-in additions, all defaulting to current behaviour so no existing
caller moves:

- `--color-outline-fill` - a resting wash under the `outline` stroke. Hover
  keeps intensifying on top of it rather than off it.
- `shape="pill"` - fully rounded ends, computed from the button's own height
  so every size lands on a true pill.
- `iconPlacement="start"` - icon ahead of the label, which centres the row
  and tightens the gap.

The app default stays `--border-radius-m`; only the error actions opt into
pills.

## Bugs found while building the playground

- **The mark squashed to an ellipse.** The card centred its children with
  `justify-content`, so content taller than it - a long stack trace, a short
  landscape viewport - shrank every flex item rather than scrolling. The 96px
  mark measured 96x48. Content now centres with `margin: auto` while the card
  scrolls, which also keeps the overflow reachable.
- **The clapperboard was drawn wrong.** Its stick crossed the board's rounded
  corner instead of hinging on it, and the rotation was negative - it swung
  further open every cycle instead of clapping shut.
- **Copy broke mid-sentence.** The body was capped at 480px, wide enough only
  for the shortest message: the unexpected-error copy split its "submit a new
  issue" link across lines, and the locked-account page read as a long line,
  an orphaned "account.", then a short line. The cap moves to 640 and the body
  balances its lines when it does have to wrap. Preformatted content opts back
  out, so a stack trace is never reflowed.

## Notes

- `routes/+error.svelte` cannot import the marks across the `_internal/`
  boundary, so its non-404 branch delegates to a new `ErrorStatusPage`, which
  also drops the two hardcoded English strings that were living in the route.
- The last commit clears two convention slips in the new code: a child
  element that carried the `trakt-` namespace a second time, and a headline
  setting `font-size` from raw `--ni-*` tokens instead of a semantic one.
- Verified in light and dark, desktop, tablet and mobile.


seferturan (approved):
Applied some fixups:

- Extracted some duplicate code.
- Removed a bunch of comments.
- Reverted the button style changes; this was introducing yet another button shape. And the latest design changes to the default button (which was merged earlier today) already fit quite nicely in this design.

PR #3229, opened 2026-09-08
feat(rating): history ratings can be changed after the fact

ElMagnea:
##Visuals

You can now **change** ratings from the history section and drill-down. 
Screenshot 2026-09-08 at 13 46 54
Screenshot 2026-09-08 at 13 46 43

## What

A rating shown in History was a dead end. Once an item was rated, the badge
sat there as a read-only tag, and the only way to change that rating was to
leave the list and open the title's summary page. Rating something and
changing that rating are the same decision, so they are now the same gesture:
the badge opens the same rate popover the empty star already opened.

Everywhere History is listed:

| Surface | Before | After |
| --- | --- | --- |
| Home / profile carousel (`PersonalHistoryList`) | star opens the popover, badge is dead | both open the popover |
| `/history` L2 (`PersonalHistoryPaginatedList`) | star opens the popover, badge is dead | both open the popover |
| `/profile/me/history` L2 (`RecentlyWatchedPaginatedList`) | no rating affordance at all | both open the popover |
| Another user's profile + L2 | read-only | unchanged, read-only |

## How it works

Rated and unrated share **one** `<Popover>` and **one** trigger:

```svelte
<Popover content={rateContent} label={rateLabel}>
  <span class="trakt-history-rating-badge" class:is-rated={Boolean(userRating)}>
    {#if userRating}<UserRating rating={userRating} />{:else}<StarIcon fill="none" />{/if}
  </span>
</Popover>
```

Two reasons for one trigger rather than two:

- **The popover survives a rating landing.** The trigger element does not
  unmount when the badge swaps in, so the stars stay open under the cursor
  and the rating can be adjusted again immediately.
- **It drops a `<button>` inside a `<button>`.** The old unrated trigger put
  an `ActionButton` inside `Popover.Trigger`, which nests two interactive
  controls and puts two stops in the tab order for one action.

Both states now carry the same hover wash and the same `:active` press. The
lone star lost `ActionButton`'s glow and gained the badge treatment - that
direction was picked from the two side by side. Negative margins cancel the
new hit-area padding, so the badge keeps the exact position it had as a tag;
only the wash grows outward.

## State invalidation

No new wiring - this already worked and is already covered. `useRatings`
fires `invalidate(InvalidateAction.Rated(type))` after both add and remove,
and `currentUserRatingsQuery` declares all four `Rated(...)` tokens in its
`invalidations`. `useUser().ratings` reads off that query, so a change made
in the carousel is reflected in the L2 view and vice versa without a reload.
Covered by `useRatings.spec.ts` ("should call invalidate after rating").

## Notes

- **`Popover` gained an optional `label`.** The rated badge's
  name-from-content is just the number - "3" - so the trigger needs an
  explicit accessible name (`button_label_change_rating`, one new key in
  `meta/en.json`; CrowdIn owns the translated catalogs). The same commit puts
  a `:focus-visible` ring back on the trigger, which `all: unset` had been
  stripping for all three consumers.
- **`isActionable` on the profile lists is a separate fix.** They passed no
  `isActionable` at all, so `/profile/me/history` offered neither a rating
  nor a row menu while `/history` offered both. They now gate on `useIsMe`,
  which keeps another user's activity read-only - a rating badge there would
  be showing *my* rating next to *their* watch. Side effect: my own profile
  L2 now also gets the row menu and the absolute date format, matching
  `/history`.
- **The rated state is auth-gated, so I could not screenshot it myself.** The
  hover and press feel were reviewed live instead.

## Verification

- `deno fmt` on the touched dirs
- `svelte-check`: 616 errors, all the pre-existing `unwasm/plugin` baseline noise, zero real
- `i18n:check`: placeholders consistent with source

### DIFF
--- a/projects/client/i18n/meta/en.json
+++ b/projects/client/i18n/meta/en.json
@@ -9410,6 +9410,15 @@
         }
       }
     },
+    "button_label_change_rating": {
+      "default": "Change rating for \"{title}\"",
+      "description": "Aria label for the badge that shows the user's own rating and, when pressed, reopens the rating picker so they can change it.",
+      "variables": {
+        "title": {
+          "type": "string"
+        }
+      }
+    },
     "button_text_watch_until_here_anchor_start": {
       "default": "Start time",
       "description": "Toggle label that treats the picked date and time as when the user started binge-watching the selected episodes."
--- a/projects/client/src/lib/components/popover/Popover.svelte
+++ b/projects/client/src/lib/components/popover/Popover.svelte
@@ -5,6 +5,7 @@
 
   type PopoverProps = {
     content: Snippet;
+    label?: string;
     open?: boolean;
     onOpenChange?: (open: boolean) => void;
   } & (
@@ -12,7 +13,8 @@
     | { children: Snippet; customAnchor?: never }
   );
 
-  const { content, open, onOpenChange, ...rest }: PopoverProps = $props();
+  const { content, label, open, onOpenChange, ...rest }: PopoverProps =
+    $props();
 
   const isControlled = $derived(open !== undefined);
 
@@ -30,7 +32,7 @@
 
 <Popover.Root bind:open={getOpen, setOpen}>
   {#if rest.children}
-    <Popover.Trigger class="trakt-popover-trigger">
+    <Popover.Trigger class="trakt-popover-trigger" aria-label={label}>
       {@render rest.children()}
     </Popover.Trigger>
   {/if}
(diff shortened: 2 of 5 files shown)


seferturan (approved):
Applied some fixups:

- Extracted as a separate component; it shares the styling with the actionbutton. Could do a follow-up here to merge them a bit more, but for now I wanted to keep the constraint that action buttons are square.
- Added the tooltip.
- The icon sizes of rated/unrated are the same now.
- Position in the footer/summary card is the same as it was before now.

PR #3230, opened 2026-09-08
refactor(rating): the star row loses its callout, its outsized star and its stroke bevel

ElMagnea:
## What

The rate row carried three things that fought the shape it is built from, and
a heart that has never been the colour it was told to be.

| | Before | After |
| --- | --- | --- |
| Star points | `stroke-linejoin: bevel` chopped all five flat | default miter — sharp |
| Selected row | last star scaled to `1.3` | every star the same size |
| Unrated row | a bold **Rate** label to the left | the stars, which already say it |
| Heart | text-coloured, shrunk to `0.8`, glowing on hover | red, 24px like the stars, washes like the badge |

## Star points

`StarIcon` drew its unfilled half as a 2px bevelled stroke. Bevelling a
five-pointed star chops every tip flat, and those hard straight edges land
exactly where the eye does — the active or hovered star, where the fill slides
over them. The join is now the default miter, which holds at these angles
(1/sin 24.75° = 2.39, inside the miter limit of 4), so the points stay sharp.

The empty star is still the outline; the fill is painted over it in the same
colour, so a rated star reads as one solid shape and both states share an outer
extent. A dimmed-fill track was tried first and rejected — it reads as a
washed-out smudge rather than a star.

## Uniform stars

`[data-highlighted]` scaled to `1.3`, so a selected row ended in one oversized
star. Emphasis is colour and opacity now. The touch-scrub rule keeps its
`translateY` lift — that is about a thumb covering the target, not about size —
and drops the scale.

## The "Rate" label

The star row is self-explanatory, and it already carried `header_rate_now` as
the group's `aria-label`, so nothing is lost for assistive tech. Removing it
left `RateNow`'s `style: 'default' | 'minimal'` prop doing nothing at all; it
and its one call site go too.

## The heart was never red

Not orange either. Every surface that hosts `RateNow` — history popover,
summary row, now-playing toast — pinned `--icon-color` on `:global(svg)`:

```scss
:global(svg) { --icon-color: var(--color-text-primary); }
```

That sets the variable on the `<svg>` itself, deeper than the wrapper Svelte
creates for the value `FavoriteButton` hands down, so the host won every time
and the heart came out text-coloured. It swallowed the old orange the same way.
The overrides are gone: what colour the heart is belongs to `FavoriteButton`,
not to its host — `components.md` bans restyling a shared component from its
consumer. The `--icon-fill-color` rules beside them go too; nothing defines
`.is-current-rating` and nothing reads that property.

The heart also stopped shrinking. `size="small"` puts `scale: 0.8` on the whole
`ActionButton`, so a 24px heart rendered at 19px beside a 24px star.

## Hover

The heart glowed where the rating badge beside it only washes. `ActionButton`
draws that glow from `--color-background-action-button`, a surface a ghost
button never paints, so it hangs under nothing. Cancelled for this button only,
matching the badge treatment from #3229 — a wash and a `0.92` press.

That glow is wrong on every ghost `ActionButton`, not just this one, but ~50
files ship them and that sweep is its own PR.

## Verification

- `deno fmt` on `projects/client/src` (bare `deno fmt` rewrites the CrowdIn
  catalogs — reverted, zero catalog changes in this PR)
- `svelte-check`: 7739 files, 0 errors, 0 warnings
- commitlint: both commits clean against `commitlint.config.js`
- Rendered through a throwaway `_design_system` route (since deleted): stars at
  none/half/full, the row at 8/10 · 7/10 · 0/10, the history popover's exact
  markup, and the heart hovered.

### DIFF
--- a/projects/client/src/lib/components/buttons/favorite/FavoriteButton.svelte
+++ b/projects/client/src/lib/components/buttons/favorite/FavoriteButton.svelte
@@ -56,7 +56,13 @@
 
 {#if style === "action"}
   <QueuedIndicator {isQueued}>
-    <ActionButton {...commonProps} {...props} style="ghost" color="default">
+    <ActionButton
+      {...commonProps}
+      {...props}
+      style="ghost"
+      color="default"
+      classList="trakt-favorite-action-button"
+    >
       <FavoriteIcon {state} --icon-color="var(--color-background-red)" />
     </ActionButton>
   </QueuedIndicator>
@@ -71,3 +77,21 @@
     {/snippet}
   </DropdownItem>
 {/if}
+
+<style lang="scss">
+  /*
+    The heart wears the same treatment as the rating badge that opens the rate
+    popover beside it: a hover wash and a press, no glow. ActionButton draws
+    that glow from --color-background-action-button, a surface a ghost button
+    never paints, so here it reads as a shadow cast by nothing. The selector
+    repeats ActionButton's own disabled guards because it has to outrank the
+    rule it cancels.
+  */
+  :global(
+    .trakt-action-button.trakt-favorite-action-button:hover:not([disabled]):not(
+        [aria-disabled="true"]
+      )
+  ) {
+    box-shadow: none;
+  }
+</style>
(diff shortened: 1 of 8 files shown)


seferturan (approved):
Applied some fixups:

- Moved the style override to the action button itself.
- The pointy stars had a very sharp look and didn't match the other icons.
- Tweaked the sizes so they'd be more aligned with the favorite icon.
- Restored the grow on hover/swipe.

PR #3231, opened 2026-09-08
refactor(button): hover is a wash or a colour, never a glow

ElMagnea:
Stacked on #3230 — **base is `feat/star-rating-cleanup`**, retarget to `main` once that lands.

## What

Two rules drew a coloured drop shadow on hover:

| Rule | Reach | Shadow |
| --- | --- | --- |
| `ActionButton :hover` | every icon button, ~50 files | `0 2px 8px -2px` from `--color-background-action-button` at 50% |
| `Button [data-style=flat]:hover` | every primary CTA | `0 4px 12px -2px` from `--color-background-button` at 45%, plus `translateY(-1px)` |

Both are mixed from the button's **own background colour**, which is why they
read as a glow rather than as elevation — a real drop shadow does not take the
hue of the thing casting it. And on a ghost button, which paints no background
at all, the glow hangs under nothing.

There is now one hover idiom everywhere, the one the favourite heart got in
#3230:

- **ghost** → washes to `--color-foreground` at 10%
- **filled** → inverts its foreground and background
- **press** → stays a `scale()`
- **nothing** casts a shadow

## Dead code that went with it

Three places existed only to cancel these shadows:

- `FavoriteButton` — the scoped `box-shadow: none` added in #3230, now covered
  by the global rule.
- `RatingList` — `:global(.ratings-drilldown:hover) { box-shadow: none }`,
  cancelling the ActionButton glow by hand.
- `card-outline-button` mixin — `transform: none` and `box-shadow: none` on
  `flat:hover`, undoing the lift and glow so a card-shaped button would sit
  still. Only the `border-color` swap was ever doing anything.

Net −51 lines.

## Not touched

The bespoke hover shadows on YIR headers, the profile all-time-review link and
a few one-off cards. They are surface decoration rather than a button idiom, and
sweeping them was explicitly out of scope.

`outline`'s `box-shadow` is untouched — that one **is** the border, not
decoration. The comment saying so has been corrected now that it no longer has
`flat`'s lift to contrast itself against.

## Verification

- `deno fmt` on `projects/client/src` (bare `deno fmt` rewrites the CrowdIn
  catalogs — zero catalog changes in this PR)
- `svelte-check`: 7739 files, 0 errors, 0 warnings
- commitlint: clean
- Rendered all four states through a throwaway `_design_system` route (since
  deleted), forcing `:hover` via CDP: ghost + filled `ActionButton`, `flat` +
  ghost `Button`. Every hovered state is a flat wash or a colour inversion, no
  shadow anywhere.

### DIFF
--- a/projects/client/src/lib/components/buttons/ActionButton.svelte
+++ b/projects/client/src/lib/components/buttons/ActionButton.svelte
@@ -162,7 +162,7 @@
 
     transition: var(--transition-increment) cubic-bezier(0.22, 1, 0.36, 1);
     transition-property:
-      background-color, color, box-shadow, transform, outline, outline-offset;
+      background-color, color, transform, outline, outline-offset;
   }
 
   @each $color in "purple", "red", "blue", "orange", "default" {
@@ -188,20 +188,8 @@
     display: none;
   }
 
-  @include for-mouse {
-    :global(#{$b}:hover#{$on}) {
-      box-shadow: 0 var(--ni-2) var(--ni-8) var(--ni-neg-2)
-        color-mix(
-          in srgb,
-          var(--color-background-action-button) 50%,
-          transparent
-        );
-    }
-  }
-
   :global(#{$b}:active#{$on}) {
     transform: scale(0.92);
-    box-shadow: none;
   }
 
   :global(#{$b}:active[disabled]),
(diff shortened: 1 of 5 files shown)


seferturan (approved):
Most of the changes were already made. I fixed up this PR to only do a dead code removal.

PR #3232, opened 2026-09-08
refactor(theme): purple takes over from the cyan accent

ElMagnea:
## What

The cyan accent — `--blue-500`, `oklch(66% 0.155 245)`, the fill on the filter sidebar's Trakt-rating slider — carried three unrelated meanings at once: *"filter is active"*, *"this is new"*, and *"focused"*. Purple now owns the first and third, so the accent reads as one system instead of three coincidences.

Every site keeps its numeric step (`blue-N` → `purple-N`).

| Surface | Change |
| --- | --- |
| Slider fill, thumb, hover | `--purple-500` / `--purple-600` — the rating bar itself |
| Valued-select border | `--purple-500` — sits right beside the slider in the filter sidebar |
| `--color-input-focus` | one line, reaching all eight focus rings |
| Calendar item indicator | both themes (see note) |
| YiR panel tint + title gradient | `--purple-500` / `--purple-800` |
| `color=\"blue\"` variant family | `--color-background-blue` → `--purple-700`, foreground → `--purple-50`, hover → `--purple-800` |
| Inline link hover | `--purple-600` / `--purple-300` |
| Dead switch tokens | `--color-tick-blue`, `--color-switch-*-blue` |
| Design-system pages | share-card movie tag, TableRow highlight |

**Calendar indicator, both themes.** Light was `--blue-700` and dark `--blue-500`. Only the dark one was strictly in scope, but moving it alone would have split the same dot across themes — purple in dark, blue in light. Both went.

## Deliberately left on blue

**The \"new\" episode/movie dot** (`MediaStatusTag`, `EpisodeStatusTag`). Cyan now means exactly one thing, and the dot sits beside a red finale and a green premiere that already read as a set.

**The `--viz-*` chart palette.** It alternates purple on odd slots and blue on even ones so adjacent series stay apart — the comment at `modes.scss:375` says so outright. A numeric-preserving swap collapses it to duplicate colours, worst in the high-contrast modes where distinguishability matters most:

| Block | Distinct after swap | Collisions |
| --- | --- | --- |
| light | 7/8 | `viz-3 == viz-8` |
| dark | 6/8 | `viz-1 == viz-4`, `viz-3 == viz-8` |
| high-contrast light | 5/8 | three pairs |
| high-contrast dark | **4/8** | every pair collides |

## Nice side effect

`--purple-500` is already `--color-ratings-trakt`. The Trakt-rating slider now paints in the Trakt-rating colour — the filter and the thing it filters finally agree.

## Follow-up debt (not in this PR)

`color=\"blue\"` on Button / ActionButton / DropdownItem now renders **purple**, duplicating `color=\"purple\"`. Only four real callers (`ConfirmationDialog`, `DeleteAccount`, `WatchlistIndicator`, `ListAction`). Collapsing the variants and deleting the dead switch tokens is a component-API change and deserves its own PR rather than riding along in a colour refactor.

## Verification

- `deno fmt --check` clean on all nine changed files
- `deno task check` → 7739 files, **0 errors, 0 warnings**
- `vitest` → **3131 passed**, 10 skipped, 0 failed
- Battery re-run after rebasing onto current `origin/main`
- Zero CrowdIn catalog files touched
- Slider, valued-select, focus ring and dot eyeballed in a throwaway design-system route in **both themes**; computed styles confirmed `oklch(56.1% 0.205 314.2)` at every swapped site in light and dark

### DIFF
--- a/projects/client/src/lib/components/dropdown/DropdownItem.svelte
+++ b/projects/client/src/lib/components/dropdown/DropdownItem.svelte
@@ -303,10 +303,10 @@
 
     &[data-color="blue"] {
       @include color(
-        var(--blue-600),
-        var(--blue-200),
-        var(--blue-600),
-        var(--blue-50)
+        var(--purple-600),
+        var(--purple-200),
+        var(--purple-600),
+        var(--purple-50)
       );
     }
 
--- a/projects/client/src/lib/components/link/Link.svelte
+++ b/projects/client/src/lib/components/link/Link.svelte
@@ -101,8 +101,8 @@
 
       @include for-mouse {
         &:hover {
-          color: var(--blue-600);
-          text-decoration-color: var(--blue-300);
+          color: var(--purple-600);
+          text-decoration-color: var(--purple-300);
         }
       }
     }
--- a/projects/client/src/lib/components/select/_internal/SelectBase.svelte
+++ b/projects/client/src/lib/components/select/_internal/SelectBase.svelte
@@ -185,7 +185,7 @@
     }
 
     &[data-has-value="true"] {
-      border-color: var(--blue-500);
+      border-color: var(--purple-500);
     }
 
     &[data-disabled] {
--- a/projects/client/src/lib/components/slider/Slider.svelte
+++ b/projects/client/src/lib/components/slider/Slider.svelte
@@ -121,7 +121,7 @@
     :global(.trakt-slider-range) {
       position: absolute;
       height: 100%;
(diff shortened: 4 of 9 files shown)


seferturan (approved):
Applied a fixup to not use purple for things that are named blue 😅

We should do a follow up here:

- Clean up palettes we do not use
- See if we want all of these actually purple, or if things still need tweaking
- Extract a bunch of these as variables

PR #3234, opened 2026-09-08
fix: stop intermittent HTTP ERROR 503 on page load

Marius-TV:
## What's happening

The app intermittently showed Chrome's `HTTP ERROR 503` page. A refresh always fixed it, and it had been getting more frequent.

The 503 was **not** coming from Cloudflare. The service worker was manufacturing it:

```ts
return new Response('Service worker fetch failed', { status: 503, ... });
```

Browsers substitute their own error interstitial for a short-bodied error status on a document request, so this read as a server outage the origin was never having.

## Why it fired

Sentry pointed at the root cause: `SecurityError: IDBFactory.open() called in an invalid security context`, 566 events across [TRAKT-WEB-B9S](https://trakt-tv.sentry.io/issues/TRAKT-WEB-B9S) and [TRAKT-WEB-9Z6](https://trakt-tv.sentry.io/issues/TRAKT-WEB-9Z6), still firing, overwhelmingly iOS.

`authMarker.ts` looked defensive but wasn't. `idb-keyval`'s `createStore` is **lazy** — it never calls `indexedDB.open()`. The open happens on first use, *synchronously*, inside `get`/`set`:

```js
const getDB = () => { ...; const request = indexedDB.open(dbName); ... };
return (txMode, callback) => getDB().then(...);   // getDB() runs synchronously
```

So in a blocked-storage context they throw **before returning a promise**, and the chained `.catch()` never attaches. `getStore()`'s try/catch was guarding a call that cannot throw.

That rejection propagates into the service worker, which reads the marker to key cached documents. A rejected cache key fails `fetchAndCachePut` and then `cacheMatch` — which Workbox does **not** wrap in a try/catch — so the strategy throws and *every* navigation falls through to the catch handler. Affected users were one network blip away from an error page on any page load. Hence: intermittent, refresh fixes it, skewed mobile.

## Changes

**`fix(auth)`** — `try`/`catch` around the `await`, which catches the synchronous throw and an async rejection alike.

**`fix(pwa)`** — two service-worker fixes:
- Fail with `Response.error()` (what the browser does with no worker installed) instead of a synthesised status. The retry is now delayed rather than running in the same tick as the fetch that just failed, so transient blips self-recover. Only GET navigations retry, so a form submission is never resent.
- Activation keeps the current build's cache. The sweep matched every name prefixed with the navigation key — its own included — and races a 3s timeout that can leave it running detached, long enough to delete documents the new build had just cached. So the fallback never survived to be used. `_cb`/CacheBust still clear everything.

Replaying another build's HTML was considered and rejected: it references `_app/immutable` chunks the deploy no longer serves, which is exactly what 8f50d7832 fixed.

## Testing

New `authMarker.spec.ts`. I verified the two "storage is blocked" cases **fail** against the old `.catch()` implementation and pass against the fix, while the async-rejection cases pass against both — so they pin the actual regression rather than passing vacuously.

Full suite: 3150 passed, 0 failures. `deno task check` and `deno task lint` show no new issues.

## Note

Production is still on `6de102491b` (2026-09-08). The `build` job in `ci_cd.yml` shares a concurrency group with `files-changed`, so rapid pushes cancel each other before the deploy step. Worth fixing separately, or this won't ship.

### DIFF
--- a/projects/client/src/lib/features/auth/authMarker.spec.ts
+++ b/projects/client/src/lib/features/auth/authMarker.spec.ts
@@ -1,0 +1,100 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+const idb = vi.hoisted(() => ({
+  createStore: vi.fn(),
+  get: vi.fn(),
+  set: vi.fn(),
+}));
+
+vi.mock('idb-keyval', () => idb);
+
+// `indexedDB` only has to exist; `createStore` is mocked, so nothing opens it.
+vi.stubGlobal('indexedDB', {});
+
+// What a storage-partitioned or locked-down context throws. `createStore` is
+// lazy, so this surfaces synchronously out of `get`/`set` rather than as a
+// rejected promise.
+const blocked = () => {
+  throw new DOMException(
+    'IDBFactory.open() called in an invalid security context',
+    'SecurityError',
+  );
+};
+
+async function importAuthMarker() {
+  vi.resetModules();
+  return await import('./authMarker.ts');
+}
+
+describe('util: authMarker', () => {
+  beforeEach(() => {
+    idb.createStore.mockReturnValue(vi.fn());
+  });
+
+  describe('when storage is blocked', () => {
+    it('should read as signed out rather than reject', async () => {
+      idb.get.mockImplementation(blocked);
+
+      const { readAuthMarker } = await importAuthMarker();
+
+      await expect(readAuthMarker()).resolves.toBe(false);
+    });
+
+    it('should discard a write rather than reject', async () => {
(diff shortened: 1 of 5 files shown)


seferturan (approved):
Added another fix to this: `createIdbPersister` had the same bug. `entries` throws synchronously too, and `persisterGc()` has no catch in `+layout.ts`, so it went out as an unhandled rejection on every cold layout load. That's [TRAKT-WEB-9Z6](https://trakt-tv.sentry.io/issues/TRAKT-WEB-9Z6) (736 events), which I'd wrongly lumped in with the authMarker one. Same try/catch treatment, plus specs.

PR #3251, opened 2026-09-10
feat(list): add featured people with a searchable drilldown

kevincador:
## Summary

Adds Featured People below the items on personal and official lists, following the cast row on media detail pages. This draft is for validating whether the feature feels useful and whether its placement, cards, and drilldown feel right.

Built on https://github.com/trakt/trakt-workers/pull/1525.

## Changes

- Treat featured people as bonus content with a generous gap after the list.
- Reuse portrait people cards and the cast row's drilldown behavior, with a searchable drawer containing all featured people.
- Preserve the API's ranking and hide the section when no featured people are available.
- Load cached person summaries as entries become visible to fill in portraits and the known-for department.

## Why this must remain a draft

The current `GET /lists/{id}?extended=featured` response contains only names and IDs for each featured person. The UI therefore makes an additional list-summary request, then a person-summary request for each visible person. Viewport gating limits the initial burst, but does not eliminate the extra requests or the delay before portraits appear.

Before this is ready to merge:

- [ ] Return compact person records with headshot URLs and relevant roles directly from the featured-people request, so the row and drawer do not need per-person summary requests. Prefer roles relevant to this list over a person's general known-for department; full biographies and credits are unnecessary.
- [ ] Consider a dedicated `GET /lists/{id}/people` endpoint for this computed collection, with independent caching and invalidation, instead of attaching it to the list summary.
- [ ] Consider exposing the appearance counts already used by the backend to rank people.
- [ ] Adapt the client to the agreed API contract and remove the temporary per-person hydration.
- [ ] Validate desktop/mobile placement, card presentation, and the view-all/search interaction; check request volume and portrait loading with the improved API.

Please keep this PR in draft while we validate the experience and resolve the API/performance work above.

## Validation

- Targeted query, list-summary, featured-people drawer, and existing cast-drawer tests: passed (8 tests).
- Svelte/TypeScript check: passed with 0 errors and 0 warnings (Node with an 8 GB heap; the initial Deno run exceeded its default heap limit).
- ESLint and formatting checks on changed code: passed; staged diff whitespace check passed.
- Live public API inspection confirmed that featured entries contain names and IDs only.
- Desktop/mobile visual verification remains pending: the local preview hit an app-wide Svelte hydration error before the list rendered. No screenshots are attached.
- Full coverage, production build, and E2E checks were not run locally. The repository's main CI and preview pipeline skips draft PRs.

## Screenshots

Screenshot 2026-09-10 at 20 53 12
Screenshot 2026-09-10 at 20 54 00

## Pull Request Checklist

Before you dim the lights and raise the curtain on your cinematic masterpiece,
make sure your pull request has passed these critical checks:

- [x] **Clear and Concise Title:** Does your title shine a spotlight on the
      essence of your changes?
- [x] **Detailed Description:** Have you set the stage with a compelling
      narrative that explains the "why" and "how" of your changes?
- [x] **Screenshots/Videos:** Have you captured the magic with visual evidence
      that showcases your changes in action?
- [x] **Tests:** Has your code undergone a rigorous dress rehearsal with unit
      tests to ensure it performs flawlessly under pressure?
- [x] **Coding Standards:** Does your code adhere to the project's style guide,
      ensuring it's as elegant as a film score?
- [x] **Commit Messages:** Are your commit messages clear and concise, guiding
      the audience through the evolution of your code?
- [ ] **Documentation:** If your changes impact the script, have you updated the
      documentation accordingly?
- [x] **Code Review:** Are you prepared to embrace feedback from the discerning
      eyes of the project maintainers and fellow contributors?

Replaces #3250 after renaming the source branch.

### DIFF
--- a/projects/client/i18n/meta/en.json
+++ b/projects/client/i18n/meta/en.json
@@ -11,6 +11,10 @@
     }
   },
   "messages": {
+    "list_title_featured_people": {
+      "default": "Featured People",
+      "description": "Heading for the row of recurring cast, directors, and creators featured in a list."
+    },
     "link_text_streaming_sync_settings": {
       "default": "Streaming",
       "description": "Text for the link to the streaming services settings page."
--- a/projects/client/src/lib/requests/queries/lists/listFeaturedPeopleQuery.spec.ts
+++ b/projects/client/src/lib/requests/queries/lists/listFeaturedPeopleQuery.spec.ts
@@ -1,0 +1,72 @@
+import { MEDIA_POSTER_PLACEHOLDER } from '$lib/utils/assets.ts';
+import { server } from '$mocks/server.ts';
+import { createTestBedQuery } from '$test/beds/query/createTestBedQuery.ts';
+import { runQuery } from '$test/beds/query/runQuery.ts';
+import { http, HttpResponse } from 'msw';
+import { describe, expect, it } from 'vitest';
+import { listFeaturedPeopleQuery } from './listFeaturedPeopleQuery.ts';
+
+const listId = 42;
+const readFeaturedPeople = () =>
+  runQuery({
+    factory: () => createTestBedQuery(listFeaturedPeopleQuery({ listId })),
+    waitFor: (response) => response.isSuccess,
+  }).then((response) => response.data);
+
+describe('listFeaturedPeopleQuery', () => {
+  it('should request the extension and preserve ranking with minimal person data', async () => {
+    let requestedUrl: URL | undefined;
+    server.use(
+      http.get(`http://localhost/lists/${listId}`, ({ request }) => {
+        requestedUrl = new URL(request.url);
+        return HttpResponse.json({
+          featured: [
+            {
+              name: 'Rebecca Ferguson',
+              ids: { trakt: 2, slug: 'rebecca-ferguson' },
+            },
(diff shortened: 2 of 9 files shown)


seferturan (changes requested):
Applied some fixups:

- Dropped the redundant wrapper div in `FeaturedPersonItem`
- Moved the featured people drawer into a typed `userListDrawerNavigation`
- `UserListDrawers` enum instead of the `"featured-people"` string literals
- Gave the preview toggle its own `preview_feature_title_featured_people` message instead of reusing the section title
- Dropped the explicit `audience="vip"` on `RenderForFeature`, it's the default already
- Dropped the `:v2` suffix from the `listFeaturedPeople` query key. Query hasn't shipped yet, so there's no stale cache to bust
- Added a 204 guard to `listFeaturedPeopleQuery`
    - A 204 is `response.ok`, so `response.json()` threw on the empty body, and the query retried 3x before settling in an error state

Furthermore, I removed the `completed` slot from `PaginatedList`. This was an abstraction leak: `PaginatedList` should just render a paginated list, what goes under it belongs in the consumer.
Which I think also indicates the biggest flaw of this change: this list is barely discoverable/easily reachable on paginated lists.

Placing the list at the top also isn't an option imo. Since the list content should be leading here, and it will take up too much space.

Maybe we add a `featured people` button somewhere to open the drawer directly? Maybe something like the activities chip we show on summary pages? Thoughts @kevincador

PR #3256, opened 2026-09-11
Show: episode count under season posters and dropdowns

rectifyer:
# Why

Resolves the Featurebase request [Episode count per season](https://roadmap.trakt.tv/p/episode-count-per-season), one of the most voted ideas on the roadmap. The previous site showed the count under every season poster and the new one dropped it.

# What

The episode count already existed in several places: the show page's season dropdown (in parentheses), the drawer's Episodes tab header, the season progress card, and the mark-as-watched confirmations. This PR adds it to the remaining places a season is named.

- Season poster cards in the seasons drawer and the mobile show page get a third footer line with the count, under the season label and its title. The footer grows one line in both title states through new `-lg` footer, card, and list height tokens, so nothing clips.
- The seasons drawer dropdown and the episode drawer's season selector now show counts. The dropdown always includes them, so the `default` / `detailed` variant prop is gone.
- Dropdown labels use the middle dot separator (`1 - Murder House · 12 eps.`) to match the related-shows footer instead of parentheses.

Every count reuses the existing `tag_text_number_of_episodes` message, so no new translations are needed.

# Screenshots

image

image

image

# Not included

Seasons inside user lists and season entries on profile ratings still show no count. Both render the show poster with a "Season 1" tag rather than a season poster, so they can follow separately if wanted.

# Validation

- `deno fmt --check`, ESLint, and svelte-check pass with no errors or warnings.
- Verified on American Horror Story in the seasons drawer, the episode drawer, the desktop show page dropdown, and the mobile poster row.

### DIFF
--- a/projects/client/src/lib/components/card/CardCover.svelte
+++ b/projects/client/src/lib/components/card/CardCover.svelte
@@ -119,6 +119,12 @@
       bottom: 0;
       inset-inline-start: 0;
 
+      /* Cards that hang indicators off the bottom edge raise the tag above them. */
+      padding-bottom: var(
+        --padding-bottom-override-card-tag,
+        var(--padding-card-tag)
+      );
+
       justify-content: flex-start;
       gap: var(--gap-xxs);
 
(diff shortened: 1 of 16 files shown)


rectifyer:
@ElMagnea Thoughts on my most recent revision using the pills? I think it works pretty well, but just doubling checking before we merge it in.

seferturan (approved):
I made some changes and reduced the scope a bit. The count in the section list were a temporary addition (as a 'quick win') to indicate that there could be more episode than there are visible on desktop, since we don't scroll there. @ElMagnea is working on a approach to completely get rid of the counts from the select.

- For counts in the drawer: I don't think it's needed there at all, the episodes section in it already states how many episodes there are in the season.
- I do like the tags more than the parenthesis; so I changed this to use the existing tags as much as possible, and kept the delta to a minimum.

